Issue #148, opened 2014-11-23
[RPMLint]   Program returns random data in a function


posophe:
I: Program returns random data in a function
E: mate-desktop no-return-in-nonvoid-function mate-color-select.c:48, 88
E: mate-desktop no-return-in-nonvoid-function test.c:56

PR #151, opened 2014-12-08
Fix missing return in non-void functions

posophe:
fix issue #148


flexiondotorg:
@stefano-k Please review this for inclusion in MATE 1.10

PR #179, opened 2015-07-03
Resurrect the old user-guide as it's complete and translated

monsta:
Big thanks to @NiceandGently for doing most of the work on this :)


infirit:
Moving discussion to here as it is the more appropriate place, see [this](https://github.com/mate-desktop/caja/pull/437#issuecomment-118060626) for some background.

There is confusion, when I say m-u-g I mean the repo with the mallard pages. __Not__ the old gfdl docbook in the help dir of applications and the user-guide dir in mate-desktop. It will be good that from now to refer to them as mallard and docbook to avoid more confusion.

>>Install m-u-g to a new directory (mate-help) and fix file collision (desktop file)

>Personally, I don't care which one is renamed as long as the users get the complete help.

The least amount of work and less confusing for user, distro maintainers and the MATE team is to just rename m-u-g, see below on mate-terminal.

>I only worried about this because of @MotoHoss. Are you sure this rename will be ok for him?

I am not sure what the issue is, did @MotoHoss raise concerns? It is just a directory name, I am sure he'll get over it :smile:.

>>Update mate-terminal help url to refect new directory

>I looked at it and saw it has its own help bundled. It just needs some small fixes.

Yes it installs gfdl docbook files but these are not used as currently the link points to the new mallard page from m-u-g. The mallard page is indeed not translated for reason mentioned before. It can be added to transifex under the condition that it will not get major reworks and will actually be __used__. It is close to zero effort to revert the url in mate-terminal to point to the old gfdl docbook page, see below for more on this.

>Won't it be better to use it instead? It's also translated - unlike the page in m-u-g.

Sure but this is part of the whole confusion. At one end ppl say they want to use the new mallard files and then I see confusion and concerns about translations and inter connectedness.. What is the goal, what is it that all this is supposed to accomplish, and what is the plan to get there? These questions are still not entirely clear to me.

The whole point of m-u-g is to get documentation under a free enough licence so debian can distribute it. But as m-u-g is nowhere near complete the proper choice should have been to just revert back to the gfdl docbook files for 1.10 and re-evaluate for 1.12.

monsta:
> I am not sure what the issue is, did @MotoHoss raise concerns? It is just a directory name, I am sure he'll get over it :smile:.

No, he didn't (we actually didn't hear from him since end of January). I only mentioned this as a possible future issue. :smile:

> It is close to zero effort to revert the url in mate-terminal to point to the old gfdl docbook page, see below for more on this.

Yes, and we intend to do that. @NiceandGently is working on that in [this branch](https://github.com/NiceandGently/mate-terminal/commits/dev-help).

> Sure but this is part of the whole confusion. At one end ppl say they want to use the new mallard files and then I see confusion and concerns about translations and inter connectedness.

You mean Debian/Ubuntu maintainers? Because I didn't hear of any *users* wishing to have the new mallard guide.

> What is the goal, what is it that all this is supposed to accomplish, and what is the plan to get there? These questions are still not entirely clear to me.

I see the goal as to bring back translated help pages for every MATE app. To accomplish this, we plan to use the old docbook help files bundled in the apps, plus the old docbook user-guide (to get help for Caja, mate-panel and general topics like using the desktop, making backups, etc.).

Where needed, the apps' code will be corrected to use the right links from the docbook guide.

NiceandGently:
A **proposal** to make all distros happy which see license probs with old docbook quide.
( i don't really care which resulting directories we use, it should be the less work for us base on what is done now)
- we ship the old docbook guide with the directory from this PR and adjust caja, mate-panel and mate-terminal for it.
-  we ship new mallard guide with latest git changes, which include help for caja, mate-terminal + some general stuff. An we a add patches for (caja+ m-t) to use this help to debian-packages repo.
In result debian can use the new mallard guide. Only help for panel is broken than for debian, but that was broken all the time.

infirit:
>I only mentioned this as a possible future issue.

Like I said, it is just the dir name, worrying about things that __may__ happen is not productive. There is a problem __now__ that needs addressing.

>I see the goal as to bring back translated help pages for every MATE app. To accomplish this, we plan to use the old docbook help files bundled in the apps, plus the old docbook user-guide (to get help for Caja, mate-panel and general topics like using the desktop, making backups, etc.).

Now that it is clear what the goals is ppl can properly review and comment on the PR's. See below for some comments.

I care not so much about the dir name (more about the plan) so feel free to pick the one. IMHO making things as they where with 1.8 is the simplest and most straightforward way of handling this but in the end it is just a dir name and if you two are going to invest significant effort I'll leave this with you to figure out.

The most complete help are the docbook files in here so we should install __this__ desktop file not the one from the m-u-g repo. The reasoning is that this is the main source for help so it should provide one. My suggestion would be to remove commit 8570e9af42344e8f87f69cea964166b764f09285 plus 0fc2aa2d104cdf26bd50a0b9cd1aea96ec9d0ed7 and disable it in the m-u-g repo instead. 

Squash the link fixes into a single commit please and then re-base against master to remove those horrible merge commits.

Last question/suggestions, did you actually talk to Mike and Vangelis about this? They are the ones most affected by all this (the name change for the repack and the mallard docs) so it would be good to get their input.

monsta:
Ok, commits are squashed and rebased.

PR #186, opened 2015-07-20
Revert "changed checking for NULL a bit"

NiceandGently:
This reverts commit 48df7a0c71a2df4050abae40ec9ca2b35dee7988.

The commit causes a regression with timing xml backgrounds.
see https://github.com/mate-desktop/mate-control-center/issues/159

### DIFF
--- a/libmate-desktop/mate-bg.c
+++ b/libmate-desktop/mate-bg.c
@@ -2857,14 +2857,9 @@ handle_text (GMarkupParseContext *context,
 	     GError             **err)
 {
 	SlideShow *parser = user_data;
+	Slide *slide = parser->slides->tail? parser->slides->tail->data : NULL;
 	FileSize *fs;
 	gint i;
-
-	g_return_if_fail (parser != NULL);
-	g_return_if_fail (parser->slides != NULL);
-	g_return_if_fail (parser->slides->tail != NULL);
-
-	Slide *slide = parser->slides->tail->data;
 
 	if (stack_is (parser, "year", "starttime", "background", NULL)) {
 		parser->start_tm.tm_year = parse_int (text) - 1900;


infirit:
Because we now return and skip the first six in the if/elseif block. We should move the second two of the g_return_if_fail down to when slide is actually used.

Does the below also fix the problem?
```diff
diff --git a/libmate-desktop/mate-bg.c b/libmate-desktop/mate-bg.c
index 8666b73..2a9f24b 100644
--- a/libmate-desktop/mate-bg.c
+++ b/libmate-desktop/mate-bg.c
@@ -2861,8 +2861,6 @@ handle_text (GMarkupParseContext *context,
 	gint i;
 
 	g_return_if_fail (parser != NULL);
-	g_return_if_fail (parser->slides != NULL);
-	g_return_if_fail (parser->slides->tail != NULL);
 
 	Slide *slide = parser->slides->tail->data;
 
@@ -2886,11 +2884,15 @@ handle_text (GMarkupParseContext *context,
 	}
 	else if (stack_is (parser, "duration", "static", "background", NULL) ||
 		 stack_is (parser, "duration", "transition", "background", NULL)) {
+
+		g_return_if_fail (parser->slides->tail != NULL);
 		slide->duration = g_strtod (text, NULL);
 		parser->total_duration += slide->duration;
 	}
 	else if (stack_is (parser, "file", "static", "background", NULL) ||
 		 stack_is (parser, "from", "transition", "background", NULL)) {
+
+		g_return_if_fail (parser->slides->tail != NULL);
 		for (i = 0; text[i]; i++) {
 			if (!g_ascii_isspace (text[i]))
 				break;
@@ -2907,12 +2909,16 @@ handle_text (GMarkupParseContext *context,
 	}
 	else if (stack_is (parser, "size", "file", "static", "background", NULL) ||
 		 stack_is (parser, "size", "from", "transition", "background", NULL)) {
+
+		g_return_if_fail (parser->slides->tail != NULL);
 		fs = slide->file1->data;
 		fs->file = g_strdup (text);
 		if (slide->file1->next != NULL)
 			parser->has_multiple_sizes = TRUE;
 	}
 	else if (stack_is (parser, "to", "transition", "background", NULL)) {
+
+		g_return_if_fail (parser->slides->tail != NULL);
 		for (i = 0; text[i]; i++) {
 			if (!g_ascii_isspace (text[i]))
 				break;
@@ -2928,6 +2934,8 @@ handle_text (GMarkupParseContext *context,
 			parser->has_multiple_sizes = TRUE;
 	}
 	else if (stack_is (parser, "size", "to", "transition", "background", NULL)) {
+		g_return_if_fail (parser->slides->tail != NULL);
+
 		fs = slide->file2->data;
 		fs->file = g_strdup (text);
 		if (slide->file2->next != NULL)
```

infirit:
Reverted for now as the above makes things worse. @monsta fyi.

monsta:
Ok, so ```slide``` **can** be allowed to be ```NULL``` in some cases... probably in these parts of this if/else spaghetti where ```slide``` itself isn't used at all. So the check should be done only in these parts, it should be enough to avoid NULL pointer dereference.

PR #187, opened 2015-07-21
bg slideshow: smarter check for NULL pointers

monsta:


NiceandGently:
Looks good with a quick test in VM, cosmos backgrounds shows me the same spiral galaxy as in 1.8 VM.

PR #190, opened 2015-07-29
Fix missing icons on GTK3 with custom icon theme

City-busz:
- Add mate icon from mate icon theme to hicolor.
- Replace help icon with the standard help-browser icon.
- Replace gcolor2 icon with the gtk-select-color (included in GTK+), which provide a more appropriate icon. Also set the window icon.


NiceandGently:
> Not really important as it comes from GTK+.

i prefer to have a full set icon theme

NiceandGently:
hmm, i do not see any icon in gtk3 rpm http://koji.fedoraproject.org/koji/rpminfo?rpmID=6566144
Which package ships the icons?

City-busz:
> Which package ships the icons?

The icons are included in libgtk as [GResources] (https://developer.gnome.org/gio/stable/GResource.html).

City-busz:
Any decision?

infirit:
I just rebased to master and merged, thanks.

raveit65:
@City-busz 
The commit breaks creating a tarball with distcheck.
```
ERROR: files left after uninstall:
./share/icons/hicolor/icon-theme.cache
Makefile:825: recipe for target 'distuninstallcheck' failed
make[1]: *** [distuninstallcheck] Error 1
make[1]: Leaving directory '/home/rave/Downloads/mate-desktop-dist/mate-desktop-1.11.0/_build'
Makefile:769: recipe for target 'distcheck' failed
```
I think the iconcache file should be excluded from distcheck

raveit65:
./autogen.sh --enable-gtk-doc --enable-deprecated --disable-strict && make -j5 && make dist -j5 && make distcheck -j5

raveit65:
I will revert the commit because we can't create 1.11 tarballs.
I'm happy to merge it again if it is fixed.
It is correct that the package don't remove the cache file from hicolors dir, but the cache file needs to exclude from distcheck.
Distcheck fails because it can't uninstall icon-theme.cache.
Another solution is to remove the creation of the cache, because this is normal the job from a scriplet in distro package.
see for example https://fedoraproject.org/wiki/Packaging:ScriptletSnippets#Icon_Cache
But as you see the cache will only created/updated.

raveit65:
update:
I removed only the iconcache generation

PR #194, opened 2015-08-19
configure.ac: drop obsolete macro

monsta:
It's been a no-op for a long time.
See https://mail.gnome.org/archives/commits-list/2011-December/msg00681.html for example.


infirit:
merged

Issue #201, opened 2015-08-31
obsolete mouse single-click gsettings key


NiceandGently:
```
[rave@mother ~]$ gsettings get org.mate.peripherals-mouse single-click
true
```
does nothing!
Caja has an own key for that, so this can savely removed.

PR #206, opened 2015-09-14
drop useless single-click key (Caja uses its own one)

monsta:
fixes https://github.com/mate-desktop/mate-desktop/issues/201


infirit:
merged

PR #209, opened 2015-11-02
refactor draw/expose code a bit and set the correct return type

monsta:


raveit65:
merged
https://github.com/mate-desktop/mate-desktop/commit/8760b08f28f44f46edcca119fef1f768a93267bd

PR #211, opened 2015-11-19
colorbutton: handle alpha for GdkRGBA

City-busz:
We have to handle alpha value to avoid uninitialized variable.

This fixes the broken color chooser in mate-panel properties on GTK3.


monsta:
Thanks, looks good here. Color select works now. (Still get applets crashing when moving transparency slider *fast* - but that's a separate issue. :smile:)

raveit65:
merged 53d0f2cd29a3ba3557f4ef4f2d7967c88a8a1914

PR #214, opened 2016-01-20
use GtkBox instead of GtkVBox in both GTK+ builds

monsta:
and fix deprecated usage of GtkHBox and GtkVBox in GTK+3 build

NOTE: this commit changes the public interface of the library

@raveit65 @lukefromdc please test for any possible issues... what might be affected are color selection dialogs in eom, mate-panel and netspeed applet. All seem to work fine here though.

### DIFF
--- a/libmate-desktop/mate-aboutdialog.c
+++ b/libmate-desktop/mate-aboutdialog.c
@@ -492,7 +492,11 @@ mate_about_dialog_init (MateAboutDialog *about)
   /* Widgets */
   gtk_widget_push_composite_child ();
 
+#if GTK_CHECK_VERSION (3, 0, 0)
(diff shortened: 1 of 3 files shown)


monsta:
Sorry, forgot to set ```orientation``` property to ```GTK_ORIENTATION_VERTICAL```. Rebased and force-pushed.

lukefromdc:
The EOM and mate-panel color selection dialogs seem fine, didn't get one
at all in netwpeed applet but never did see one there. The applet seems to
be fine.

monsta:
In netspeed applet it's in Device Details menu item - there will be a dialog with two color buttons.

lukefromdc:
OK, got the color buttons in netspeed, the color chooser widget works fine.

PR #230, opened 2016-06-16
port GdkColors to GdkRGBA

raveit65:
For this change there is some more work in other packages needed.
```
[rave@mother github-matedesktop]$ grep -nr libmate-desktop/mate-bg *
caja/libcaja-private/caja-directory-background.c:42:#include <libmate-desktop/mate-bg.h>
caja/src/caja-application.c:78:#include <libmate-desktop/mate-bg.h>
caja/eel/eel-background.c:43:#include <libmate-desktop/mate-bg.h>
caja/eel/eel-background.h:51:#include <libmate-desktop/mate-bg.h>
mate-control-center/capplets/appearance/mate-wp-item.h:26:#include <libmate-desktop/mate-bg.h>
mate-control-center/capplets/appearance/appearance-desktop.c:31:#include <libmate-desktop/mate-bg.h>
mate-panel/mate-panel/panel-background-monitor.c:36:#include <libmate-desktop/mate-bg.h>
mate-screensaver/src/gs-manager.c:32:#include <libmate-desktop/mate-bg.h>
mate-settings-daemon/plugins/background/msd-background-manager.c:43:#include <libmate-desktop/mate-bg.h>
```
But this is the initial change and mainly discussions should be done here.
Note, for testing you need also apply the  https://github.com/mate-desktop/caja/tree/dev-style-context branch, otherwise you will get the caja spanning window issue until you disable 'show desktop icons'.
This means we need to push this all together.
@monsta @dnk @flexiondotorg @lukefromdc @XRevan86
Review, testing and help is appreciated

Thanks to @XRevan86 who did help me already

### DIFF
--- a/libmate-desktop/mate-bg-crossfade.c
+++ b/libmate-desktop/mate-bg-crossfade.c
@@ -273,9 +273,21 @@ tile_surface (cairo_surface_t *surface,
 	}
 	else
 	{
+#if GTK_CHECK_VERSION (3, 0, 0)
+		GtkStyleContext *context;
+		GdkRGBA bg;
+		context = gtk_style_context_new ();
+		gtk_style_context_add_provider (context,
+										GTK_STYLE_PROVIDER (gtk_css_provider_get_default ()),
+										GTK_STYLE_PROVIDER_PRIORITY_THEME);
+		gtk_style_context_get_background_color (context, GTK_STATE_FLAG_NORMAL, &bg);
+		gdk_cairo_set_source_rgba(cr, &bg);
+		g_object_unref (G_OBJECT (context));
+#else
 		GtkStyle *style;
 		style = gtk_widget_get_default_style ();
 		gdk_cairo_set_source_color(cr, &style->bg[GTK_STATE_NORMAL]);
+#endif
 	}
 
 	cairo_paint (cr);
--- a/libmate-desktop/mate-bg.c
+++ b/libmate-desktop/mate-bg.c
@@ -1306,12 +1306,13 @@ mate_bg_is_dark (MateBG *bg,
 	if (pixbuf) {
 #if GTK_CHECK_VERSION (3, 0, 0)
 		GdkRGBA argb;
+		guchar a, r, g, b;
 
 		pixbuf_average_value (pixbuf, &argb);
-		guchar a = argb.alpha * 0xff;
-		guchar r = argb.red * 0xff;
-		guchar g = argb.green * 0xff;
-		guchar b = argb.blue * 0xff;
+		a = argb.alpha * 0xff;
+		r = argb.red * 0xff;
+		g = argb.green * 0xff;
+		b = argb.blue * 0xff;
 #else
 		guint32 argb = pixbuf_average_value (pixbuf);
 		guchar a = (argb >> 24) & 0xff;
--- a/libmate-desktop/mate-bg.c
+++ b/libmate-desktop/mate-bg.c
@@ -101,8 +101,13 @@ struct _MateBG {
 	char		*filename;
 	MateBGPlacement	 placement;
 	MateBGColorType	 color_type;
+#if GTK_CHECK_VERSION (3, 0, 0)
+	GdkRGBA	 	 primary;
+	GdkRGBA	 	 secondary;
+#else
 	GdkColor	 primary;
 	GdkColor	 secondary;
+#endif
 	gboolean	 is_enabled;
 
 	GFileMonitor* file_monitor;
@@ -153,8 +158,13 @@ static GdkPixbuf *pixbuf_scale_to_min  (GdkPixbuf  *src,
 
 static void       pixbuf_draw_gradient (GdkPixbuf    *pixbuf,
 					gboolean      horizontal,
+#if GTK_CHECK_VERSION (3, 0, 0)
+					GdkRGBA     *c1,
+					GdkRGBA     *c2,
+#else
 					GdkColor     *c1,
 					GdkColor     *c2,
+#endif
 					GdkRectangle *rect);
 
 static void       pixbuf_tile          (GdkPixbuf  *src,
@@ -206,12 +216,35 @@ static FileSize   *find_best_size      (GSList                *sizes,
 					gint                   width,
 					gint                   height);
 
+#if GTK_CHECK_VERSION (3, 0, 0)
+static void
+color_from_string (const char *string,
+		   GdkRGBA   *colorp)
+{
+	/* If all else fails use black */
+	gdk_rgba_parse (colorp, "#000000");
+
+	if (!string)
+		return;
+
+	gdk_rgba_parse (colorp, string);
+}
+
+static char *
+color_to_string (const GdkRGBA *color)
+{
+	return g_strdup_printf ("#%02x%02x%02x",
+				((guint) (color->red * 65535)) >> 8,
+				((guint) (color->green * 65535)) >> 8,
+				((guint) (color->blue * 65535)) >> 8);
+}
+#else
 static void
 color_from_string (const char *string,
 		   GdkColor   *colorp)
 {
 	/* If all else fails use black */
-	gdk_color_parse ("black", colorp);
+	gdk_color_parse ("#000000", colorp);
 
 	if (!string)
 		return;
@@ -227,6 +260,7 @@ color_to_string (const GdkColor *color)
 				color->green >> 8,
 				color->blue >> 8);
 }
+#endif
 
 static gboolean
 do_changed (MateBG *bg)
@@ -345,7 +379,11 @@ mate_bg_load_from_gsettings (MateBG    *bg,
 	char    *tmp;
 	char    *filename;
 	MateBGColorType ctype;
+#if GTK_CHECK_VERSION (3, 0, 0)
+	GdkRGBA c1, c2;
+#else
 	GdkColor c1, c2;
+#endif
 	MateBGPlacement placement;
 
 	g_return_if_fail (MATE_IS_BG (bg));
@@ -533,15 +571,25 @@ mate_bg_new (void)
 void
 mate_bg_set_color (MateBG *bg,
 		    MateBGColorType type,
+#if GTK_CHECK_VERSION (3, 0, 0)
+		    GdkRGBA *primary,
+		    GdkRGBA *secondary)
+#else
 		    GdkColor *primary,
 		    GdkColor *secondary)
+#endif
 {
 	g_return_if_fail (bg != NULL);
 	g_return_if_fail (primary != NULL);
 
 	if (bg->color_type != type			||
+#if GTK_CHECK_VERSION (3, 0, 0)
+	    !gdk_rgba_equal (&bg->primary, primary)			||
+	    (secondary && !gdk_rgba_equal (&bg->secondary, secondary))) {
+#else
 	    !gdk_color_equal (&bg->primary, primary)			||
 	    (secondary && !gdk_color_equal (&bg->secondary, secondary))) {
+#endif
 
 		bg->color_type = type;
 		bg->primary = *primary;
@@ -577,8 +625,13 @@ mate_bg_get_placement (MateBG *bg)
 void
 mate_bg_get_color (MateBG		*bg,
 		   MateBGColorType	*type,
+#if GTK_CHECK_VERSION (3, 0, 0)
+		   GdkRGBA		*primary,
+		   GdkRGBA		*secondary)
+#else
 		   GdkColor		*primary,
 		   GdkColor		*secondary)
+#endif
 {
 	g_return_if_fail (bg != NULL);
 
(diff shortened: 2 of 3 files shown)


lukefromdc:
OK, just tested this with the latest Caja changes (GTK+-3 eel: rename GdkRGBA functions), same results as last night. The issue with the "information" panel can be fixed in themes or better with a style class following .view added to it in Caja's code, I have been using it as a guinea pig for the overall BG setting functions though and have thus deliberately not fixed it in such ways.  Within Caja itself that leaves the gradient backgrounds to fix-and the long-broken setting of custom folder backgrounds.

Probably I will try to fix that info panel myself today, since I now hear the GTK3 port was never finished and it may thus not be a valid test case for eel-background.c after this porting.

lukefromdc:
I just tested the dev-colors branch of mate-control-center, the desktop background icon for a solid color background still renders black, though a solid color desktop renders correctly. Since gradient backgrounds render black on the desktop itself, could these problems be related. maybe from the same file?

raveit65:
I think so.....
appearance-desktop.c from m-c-c needs to be ported to GtkStyleContext, but i don't think this is related to this regression.
I think it is related to the port here because i see no errors in dev-colors branch from m-c-c, and the regression got pulls in after changes here,  but i'm not sure.
@dnk @XRevan86
Can you help please?

lukefromdc:
That's supposed to have been ported in the dev-colors branch of m-c-c
but does not in fact actually work for showing the desktop color background.
Image backgrounds show up fine in the background selector in m-c-c with or
without these changes.

raveit65:
I think the function is here
https://github.com/mate-desktop/mate-desktop/blob/master/libmate-desktop/mate-bg.c#L806

lukefromdc:
Just tested this, no change in Caja behavior. Gradient backgrounds, m-c-c icon for color background, info panel still render black.

lukefromdc:
Just submitted a PR tp Caja's dev-style-context branch (only one I can build/run right now) to fix the Caja information panel issue. Very simple: puts the .view style class on it, excludes rest of caja_information_panel_update_appearance used in GTK 2 to set the background in a GTK3 build to stop segfault on dragging background to info panel. Works with all themes. Can even accept a dragged background from "backgrounds and emblems" though it won't be saved or appear in another window. Code for saving that is either in the excluded code that segfaulted or in broken code elsewhere.

lukefromdc:
Just found the cause of that ugly mess with GTK 3.21.3 where the desktop does not redraw:
https://github.com/mate-desktop/caja/issues/575
Caused by
https://git.gnome.org/browse/gtk+/commit/?id=2c7b21718f18ab743cd79784ae28cad582f58503

lukefromdc:
More experiments with Caja and GTK 3.21/3 tonight.

killing mate-settings-daemon or (in another DE) starting mate-settings-daemon, cinnamon-settings-daemon, or gnome-settings-daemon will force the background to redraw and "clean" the desktop. At this point icons and text render normally. Clicking on the desktop will darken shadows until after about 4 clicks they are opaque. This or any other interaction redraws all icons and text without erasing.  Behavior is as though on each redraw the entire desktop canvas was composited with the last copy of itself rather than the background. The failed redraw also blocks background changes, though that will work with icons on the desktop turned off.

lukefromdc:
I had no luck playing with the three functions recently listed, but when I decided to try replacing them with their GNOME equivalents I found something interesting: GNOME never ported this to GdkRBGA themeslves, in fact they have not touched  gnome-desktop/libgnome-desktop/gnome-bg.c in TWO YEARS:

https://github.com/GNOME/gnome-desktop/blob/master/libgnome-desktop/gnome-bg.c

Since SOMETHING renders the background in a GNOME session, I assume this means they are still actually using this GdkColor code (long deprecated) themselves so so can we.  Cinnamon too is still using GdkRBGA in this. GTK 3.21.3 seems to work fine in GNOME, so unless we find a fix for the backgrounds by chance I think we can simply wait for GNOME to update this and then copy what they do so long as it isn't removal of color and gradient backgrounds.  Here's GNOME's "struct" declaration for the whole GNOME bg, with GdkColor right there:

struct _GnomeBG
{
	GObject                 parent_instance;
	char *			filename;
	GDesktopBackgroundStyle	placement;
	GDesktopBackgroundShading	color_type;
	GdkColor		primary;
	GdkColor		secondary;

	GFileMonitor *		file_monitor;

	guint                   changed_id;
	guint                   transitioned_id;
	guint                   blow_caches_id;

	/* Cached information, only access through cache accessor functions */
        GnomeBGSlideShow *	slideshow;
	time_t			file_mtime;
	GdkPixbuf *		pixbuf_cache;
	int			timeout_id;

	GList *		        file_cache;
`};`

More GdkColor throughout this file in GNOME and in Cinnamon. Nobody has ported this to GdkRBGA yet, I'm rolling these three packages back to Master where I am and suggest closing these PRs for now

raveit65:
Gnome-shell do not use this files any more to write the background, for this reason it was never ported :smile: 
It was used by former fallback modus.
All i know is that the background is handled by gnome-settings-daemon now.

lukefromdc:
Where does gnome-settings-daemon get the background? If they 
reimplemented it in g-s-d did they keep gradient and color BG code?

raveit65:
see, https://git.gnome.org/browse/gnome-settings-daemon/commit/?id=99da10fa1c771bd692bf00b6c4ccf2a5c1818bdb
it is moved to the shell itself.
I will ask M.Classen about, but i'm sure that gnome-bg.c isn't used anymore for gnome-shell.

raveit65:
Ok, i catched a gnome guy at irc
```
<raveit65> halfline, another question, does gnome-shell use bg code from gnome-desktop  gnome-bg.c ?
<halfline> only a small part of it
<halfline> the rendering bits are implemented in the shell itself
<halfline> but the slideshow parser is still used
<raveit65> I'm still wondering that it isn't ported to GdkRGBA
<halfline> raveit65: you mean gnome_bg_set_color ?
<halfline> probably could be changed
<raveit65> halfline, good to know, we're switching to GdkRGBA for mate and someone ask me why gnome didn't do the port
<halfline> would need to fix all the users of it
<halfline> raveit65: i would say the biggest reason it wasn't changed is we try to touch that library as little as possible
<raveit65> halfline, yes, exactly, what was used to draw th bg desktop in gnome2
<halfline> oh you mean the implementation, not just the api
<halfline> yea there's absolutely not good reason it still uses GdkColor in the implementation
```

lukefromdc:
In tonight's experiments with dev-colors and the gradient isssue (gtk rolled back to 3.21.2 before the no-redraw bug) I got no new results but found this function in mate-bg.c, parsing the pair of GkdRBGA's exactly as though they were GdkColors:

```
static guchar *
#if GTK_CHECK_VERSION (3, 0, 0)
create_gradient (const GdkRGBA *primary,
		 const GdkRGBA *secondary,
#else
create_gradient (const GdkColor *primary,
		 const GdkColor *secondary,
#endif
		 int	         n_pixels)
{
	guchar *result = g_malloc (n_pixels * 3);
	int i;

	for (i = 0; i < n_pixels; ++i) {
		double ratio = (i + 0.5) / n_pixels;

		result[3 * i + 0] = ((guint16) (primary->red * (1 - ratio) + secondary->red * ratio)) >> 8;
		result[3 * i + 1] = ((guint16) (primary->green * (1 - ratio) + secondary->green * ratio)) >> 8;
		result[3 * i + 2] = ((guint16) (primary->blue * (1 - ratio) + secondary->blue * ratio)) >> 8;
	}

	return result;
}

```
Unless the internal structure of a GdkRBGA and of a GdkColor are the same I don't see how this could work.

monsta:
So looks like only gnome-flashback still uses that part of library?

raveit65:
> So looks like only gnome-flashback still uses that part of library?

Not sure, as flashback have to use g-s-d from gnome-shell.

XRevan86:
```c
static guchar *
create_gradient (const GdkRGBA *primary,
                 const GdkRGBA *secondary,
                 int            n_pixels)
{
    guchar *result = g_malloc (n_pixels * 3);
    int i;

    for (i = 0; i < n_pixels; ++i) {
        double ratio = (i + 0.5) / n_pixels;

        result[3 * i + 0] = (guchar) ((primary->red * (1 - ratio) + secondary->red * ratio) * 0x100);
        result[3 * i + 1] = (guchar) ((primary->green * (1 - ratio) + secondary->green * ratio) * 0x100);
        result[3 * i + 2] = (guchar) ((primary->blue * (1 - ratio) + secondary->blue * ratio) * 0x100);
    }

    return result;
}
```

raveit65:
Last commit fixes the issue with gradiant colors :)

Thanks @lukefromdc for the finder
Thanks @XRevan86 for fixing that

Ok guys, now someone needs to do a code review for all 3 PRs, maybe i dis a typo somewhere.

lukefromdc:
All backgrounds (solid, gradient, and image) now render properly in Caja's dev-style-context version with this version of mate-desktop. If icons on the desktop are turned OFF, a solid color background will render black, gradients work, images work. This matches the behavior in mate-appearance-properties (dev-colors version) background where gradients show correctly but a solid color renders black. Probably the "background" plugin in mate-settings-daemon is to blame.

raveit65:
sigh, i didn't test this case.
But it's only the preview in m-c-c which render black with a solid color.
The desktop itself shows me the right solid color if caja doesn't handle the desktop.

lukefromdc:
I will look in a  few minutes for another occurance of the sort of error that just got fixed, as soon as I finish this work on a theme bug

lukefromdc:
Behavior turned out to invoke a Compiz bug in the solid background 
case-and this bug is pre-existing and affected my work with the new
GTK 3.21.3 versions and my "caja-desktop" nautius-desktop fork as 
well.

At the moment, with all the  style context changes applied and all the 
other deprecation PR's combined together on Caja, if I switch to Marco 
I can turn icons on and off with a solid background and it works. 
Still shows black in m-s-d but works.

If I switch back to Compiz I get inconsistant behavior. All the gradients
and images work, but I get redraw failures with a solid BG. A solid BG,
icons, and compiz work fine. Turn the icons off and they still show but are
no longer responsive (they are not REALLY there).  Sometimes m-s-d
now locks into that background and has to be restarted to change it again.
Sometimes I've seen black backgrounds, but often just the LAST background
shown. 

The other Caja PR's should not affect this as Caja isn't handling the desktop
in this case.  I'ts nasty but I think this is some kind of compiz corner case,
and compiz has had a buggy rep since the beginning.

monsta:
All three PRs work fine here, with both GTK+ builds.

raveit65:
After testing again it shows me that it doesn't make any different if caja handle the desktop or m-s-d. Which makes sense.
The thumbnail creation fails with solid bg color.
Thumbnails are created with libmate-desktop/mate-desktop-thumbnail.c and stored in ~.cache/thumbnails/normal.
I noticed also that if you set another color with colorchooser the color is now stored as rgba value as gsettings key in dconf-editor.
My theory is that the tumbnailer can read the gradient value but not the rgba value from solid color.
I will try to revert the port GtkColorButton to GtkColorChooser in m-c-c.

monsta:
Maybe Alberts could help us here?

lukefromdc:
OK, here's what I just found by following the code starting from the term "thumbnail:"

wp_update_preview  Calls mate_desktop_thumbnail_factory_generate_thumbnail from mate-desktop-thumbnailer.c to return a pixbuf. This appears to be the code that calls all the external thumbnailers for video, jpeg/png images etc and falls back to  gdk-pixbuf when nothing else works.

I'm guessing that this code is getting sent an image from image backgrounds and that a gradient is also an image, even if made from two identical colors. Thus those all work. When a solid BG is selected, perhaps mate_desktop_thumbnail_factory_generate_thumbnail
is being asked to thumbnail a color instead of an image and breaking?

lukefromdc:
OK, here's what I just found by following the code starting from the term "thumbnail:"

wp_update_preview  in mate-control-center/capplets/appearance/appearance-desktop.c  calls mate_desktop_thumbnail_factory_generate_thumbnail from mate-desktop-thumbnailer.c to return a pixbuf. This appears to be the code that calls all the external thumbnailers for video, jpeg/png images etc and falls back to  gdk-pixbuf when nothing else works.

"reload item" in mate-control-center/capplets/appearance/appearance-desktop.c calls mate_wp_item_get_thumbnail in mate-control-center/capplets/appearance/appearance-desktop.c/mate-wp-item.c , which in turn calls MateDesktopThumbnailFactory in mate-desktop/libmate-desktop/mate-desktop/thumbnail.c

I'm guessing that something in there is getting sent an image from image backgrounds and that a gradient is also an image, even if made from two identical colors. Thus those all work. When a solid BG is selected, perhaps mate_desktop_thumbnail_factory_generate_thumbnail
is being asked to thumbnail a color instead of an image and breaking?

PR #233, opened 2016-08-01
Reduce the number of Xrandr related probes being generated when there…

pyhalov:
… is a change in the number of screens, etc


pyhalov:
Another patch from Sun JDS

monsta:
That patch predates upstream fixes, so we already have the necessary code.

https://git.gnome.org/browse/gnome-desktop/commit/?id=d9725187de6718dd998ffa44cd15ea1bcea49340
https://git.gnome.org/browse/gnome-desktop/commit/?id=778fcb5b9e9202e5f95b6e8a22074c56d852269a

Issue #237, opened 2016-10-15
black caja filename color in icon view with dark background and gtk2


nokangaroo:
The font color of the filenames in icon view does not change to white with a dark caja
background color. The gtk2 version requires a compilation check for "intensity". I 
originally posted this in the caja issues but it belongs here. A more correct form of
the patch posted there would be something like
```
--- mate-desktop-1.16.0/libmate-desktop/mate-bg.c	2016-07-02 10:54:50.000000000 +0200
+++ mate-desktop-1.16.0/libmate-desktop/mate-bg.c	2016-10-14 18:43:00.174255800 +0200
@@ -1327,9 +1327,15 @@
 		g_object_unref (pixbuf);
 	}
 
+#if GTK_CHECK_VERSION (3, 0, 0)
 	intensity = ((guint) (color.red * 65535) * 77 +
-		     (guint) (color.green * 65535) * 150 +
-		     (guint) (color.blue * 65535) * 28) >> 16;
+			(guint) (color.green * 65535) * 150 +
+			(guint) (color.blue * 65535) * 28) >> 16;
+#else
+	intensity = (color.red * 77 +
+			color.green * 150 +
+			color.blue * 28) >> 16;
+#endif
 
 	return intensity < 160; /* biased slightly to be dark */
 }
```
which should not break the gtk3 build. (The original file seems to mix spaces and tabs).

raveit65:
The way to bring in changes is to do a pull request.
Thank you.

raveit65:
Btw. this breaks fg color on light backgrounds.

PR #238, opened 2016-10-17
GTK+-2: fix fontcolor for dark bg in caja

raveit65:
Fixes https://github.com/mate-desktop/mate-desktop/issues/237
Another gtk2 regression with caja.
This fixes the issue for me with using a dark gtk theme and setting a dark bg via 'Emplems and Backgrounds'.
Tested on top of 1.16, hope we don't get any fallout in other places.
@monsta 
Please test

### DIFF
--- a/libmate-desktop/mate-bg.c
+++ b/libmate-desktop/mate-bg.c
@@ -1327,9 +1327,15 @@ mate_bg_is_dark (MateBG *bg,
 		g_object_unref (pixbuf);
 	}
 
+#if GTK_CHECK_VERSION (3, 0, 0)
 	intensity = ((guint) (color.red * 65535) * 77 +
 		     (guint) (color.green * 65535) * 150 +
 		     (guint) (color.blue * 65535) * 28) >> 16;
+#else
+	intensity = (color.red * 77 +
+		     color.green * 150 +
+		     color.blue * 28) >> 16;
+#endif
 
 	return intensity < 160; /* biased slightly to be dark */
 }


monsta:
Confirmed, now the color is properly chosen depending on light/dark background. BTW, GTK+ theme doesn't seem to apply here as I'm using a light theme (Mint-X).

Looks like it was a regression in https://github.com/mate-desktop/mate-desktop/commit/b9b4ff6d1e0b977ce815561f1b8ea16f7ce16aec. How did I miss that...

raveit65:
I tested it with switching from our light themes to BlackMate and back, this is working here.

monsta:
Ah right, I forgot to test that. Yes, it works here too.

Found one more issue though... opening a new tab when bg is dark makes font color on that tab dark as well... :confused:

monsta:
Oh, that's already mentioned in https://github.com/mate-desktop/caja/issues/643. Might be a separate issue, I think I've seen it already in previous releases.

raveit65:
Why did this guy double posting?
....strictly forbidden :)
Ok, i will merge this because it fixes the issue in most cases, and cherry-pick it to 1.16.
Looks like it is time for a point 1 release for caja and mate-desktop, or are they another fixes?
I recall that someone had fix the issue with the second tab in the past.

monsta:
Yes, we can merge it and pick for 1.16 too. Not sure if there are any more regressions and fixes for them :smile:

PR #244, opened 2017-01-17
Drop mpaste.

flexiondotorg:
As discussed in `#mate-dev` a couple of weeks ago, this pull request drop `mpaste` from `mate-desktop` because we will not be reinstating the paste micro-site on http://mate-desktop.org.


monsta:
Works ok here. This also allows us to drop build-deps and runtime deps on Python packages - if I didn't miss anything, there are no more Python bits in this project after this PR is applied.

Issue #170, opened 2015-05-09
mate-bg.c find_best_size can return NULL, causes caja SEGV


matobinder:
So I've been setting up some VMs running TIgerVNC Server, and VNC'ing into them with TigerVNCViewer.  I'm not sure why, but when when you have a desktop background wallpaper, resizing the desktop causes Caja to crash. I looked into this and I see that mate-bg.c:find_best_size() can, and does return NULL if it doesn't find  a "best".  In my case this cases get_pixbuf_for_size to SEGV, because it tries to defreference that NULL poninter.  In all uses of find_best_size, it looks like it assumes that it will NOT return NULL, so the crash can occur in other instances as well.

I was able to stop Caja from crashing, by not having a background image... i.e.

dconf write /org/mate/desktop/background/picture-filename "''"

So, I think we need to fix find_best_size() to make sure it returns something. Instead of NULL. I don't have a good dev environment to test anything so I'm not comfortable trying to make a code change/pull request with out really trying this out myself.

But I would suggest the following... code snippt fom mate-bg.c:find_best_size()

....
for (pass = 0; pass < 2; pass++) {
        for (s = sizes; s; s = s->next) {
        FileSize *size = s->data;

```
    //here's my suggested fix... so best ALWAYS at least gets assigned once.
    if  (best == NULL) {
         best = size;
     }
    //end my fix

    i f (pass == 0 && (size->width < width || size->height < height))
           continue;
```

....
##

olivergg:
I have the same issue on an ArchLinux machine. I had to add X-MATE-AutoRestart=false to prevent the infinite loop.
Now, I can log in... but I can't start the file manager anymore : 
I've got the following error in dmesg : 
``` caja[702]: segfault at 125e84 ip b6b76c00 sp bfc355f0 error 4 in libgobject-2.0.so.0.4400.1[b6b45000+5a000]```
Can you confirm this is the same issue (segfault in libgobject) ?

raveit65:
@matobinder 
I can do a sratch koji build for epel7 with your fix for testing.
I do not use vnc, so i am not able to confirm if this really works.

monsta:
The question is, whether the patch is just a suggestion for us, or did @matobinder actually tested it?

[...]

PR #245, opened 2017-01-17
Fix to prevent find_best_size() returning NULL.

flexiondotorg:
Simple fix based on feedback in #170.

@monsta @raveit65 Please review.


raveit65:
@flexiondotorg 
Does this fix the crash if you use vnc to connect to a mate session?

PR #249, opened 2017-02-02
Fixes for background drawing

XRevan86:
This fixes directory background drawing in #506, fixes compiz-reloaded/compiz#40 (desktop background issues with Compiz 0.8.x) and fixes background fading.
Caja now draws the desktop background again and the gap in code between GTK+ 3.22 and previous versions is removed.
Other parts of the change: mate-desktop/caja#731, mate-desktop/mate-settings-daemon#169.

### DIFF
--- a/libmate-desktop/mate-bg-crossfade.c
+++ b/libmate-desktop/mate-bg-crossfade.c
@@ -41,9 +41,11 @@
 struct _MateBGCrossfadePrivate
 {
 	GdkWindow       *window;
+	GtkWidget       *widget;
 	int              width;
 	int              height;
 	cairo_surface_t *fading_surface;
+	cairo_surface_t *start_surface;
 	cairo_surface_t *end_surface;
 	gdouble          start_time;
 	gdouble          total_duration;
@@ -136,6 +138,11 @@ mate_bg_crossfade_finalize (GObject *object)
 		fade->priv->fading_surface = NULL;
 	}
 
+	if (fade->priv->start_surface != NULL) {
+		cairo_surface_destroy (fade->priv->start_surface);
+		fade->priv->start_surface = NULL;
+	}
+
 	if (fade->priv->end_surface != NULL) {
 		cairo_surface_destroy (fade->priv->end_surface);
 		fade->priv->end_surface = NULL;
@@ -203,7 +210,10 @@ mate_bg_crossfade_init (MateBGCrossfade *fade)
 {
 	fade->priv = MATE_BG_CROSSFADE_GET_PRIVATE (fade);
 
+	fade->priv->window = NULL;
+	fade->priv->widget = NULL;
 	fade->priv->fading_surface = NULL;
+	fade->priv->start_surface = NULL;
 	fade->priv->end_surface = NULL;
 	fade->priv->timeout_id = 0;
 }
@@ -303,17 +313,17 @@ mate_bg_crossfade_set_start_surface (MateBGCrossfade* fade, cairo_surface_t *sur
 {
 	g_return_val_if_fail (MATE_IS_BG_CROSSFADE (fade), FALSE);
 
-	if (fade->priv->fading_surface != NULL)
+	if (fade->priv->start_surface != NULL)
 	{
-		cairo_surface_destroy (fade->priv->fading_surface);
-		fade->priv->fading_surface = NULL;
+		cairo_surface_destroy (fade->priv->start_surface);
+		fade->priv->start_surface = NULL;
 	}
 
-	fade->priv->fading_surface = tile_surface (surface,
-						   fade->priv->width,
-						   fade->priv->height);
+	fade->priv->start_surface = tile_surface (surface,
+						  fade->priv->width,
+						  fade->priv->height);
 
-	return fade->priv->fading_surface != NULL;
+	return fade->priv->start_surface != NULL;
 }
 
 static gdouble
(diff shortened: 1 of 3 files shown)


raveit65:
Seems to work fine :-)
I tested all 3 PRs with f24 (gtk+-3.20) and fedora rawhide (f26) with gtk+-3.22.
As fedora rawhide is quite buggy in this early state i back ported the changes to 1.16 and created branches for every package. Than i can test this on stable f25 with 1.16.x, also i can ask some guys in fedora-forum to test all with compiz.
I am not sure if we want to merge it to 1.16 too as it is a massive change, but i really need patches for 1.16 to fix broken compiz in f25.
No need let the users wait 6 months for a fix which comes regular with mate-1.18 on f26.

lukefromdc:
This one should probably be merged now. It fixes the background-fade issue and causes no problems with Caja from current GIT master. Yes, the held-up Caja changes depend on this PR, but this PR does not depend on changes in caja nor in mate-settings-daemon. Ready to go

raveit65:
I think merging only thes PR is only possible for gtk+-3.22 but not for older versions.
Maybe @XRevan86 can explain this better.

lukefromdc:
We need to get this merged so that https://github.com/mate-desktop/caja/pull/731 can be merged. I can't merge that one until this one gets merged because the Caja changes depend on it.

raveit65:
I am just wating for @monsta as this huge change should be reviewed by another reviewer.

lukefromdc:
Meanwhile, I've been using this every day with no adverse issues. Good to wait for another review on another system though

libmate-desktop/mate-bg-crossfade.c
```diff
@@ -507,18 +631,45 @@ on_finished (MateBGCrossfade *fade)
...
+		if (fade->priv->widget != NULL) {
+			g_signal_connect (fade->priv->widget, "draw",
+			                  (GCallback) on_widget_draw, fade);
+		}
+	} else {
+		cairo_t   *cr;
+		GdkDisplay *display = gdk_window_get_display (fade->priv->window);
+
+		fade->priv->fading_surface = get_root_pixmap_id_surface (display);
+		g_assert (fade->priv->fading_surface != NULL);
```
monsta: Hmm... ```get_root_pixmap_id_surface``` can return NULL - should we really use an assert here?

libmate-desktop/mate-bg-crossfade.c
```diff
@@ -528,18 +679,37 @@ mate_bg_crossfade_start (MateBGCrossfade *fade,
...
 }
 
+/**
+ * mate_bg_crossfade_start_widget:
+ * @fade: a #MateBGCrossfade
+ * @widget: The #GtkWidget to draw crossfade on
+ *
+ * This function initiates a quick crossfade between two surfaces on
+ * the background of @widget.  Before initiating the crossfade both
+ * mate_bg_crossfade_start() and mate_bg_crossfade_end() need to
```
monsta: Minor issue: looks like ```mate_bg_crossfade_set_start_surface``` and ```mate_bg_crossfade_set_end_surface``` should be meant here instead? (Same as in the function above - where it was copied from)

monsta:
> This one should probably be merged now. It fixes the background-fade issue and causes no problems with Caja from current GIT master. Yes, the held-up Caja changes depend on this PR, but this PR does not depend on changes in caja nor in mate-settings-daemon. Ready to go

Existing API functions are the same, but ac697034625703b59ee4f927c9d1525b90c677ae warns about "breaking API behavior".
Indeed, I get Caja from git master crashing right after start with this PR. Guess it should be so?

```
(caja:18116): Gdk-ERROR **: The program 'caja' received an X Window System error.
This probably reflects a bug in the program.
The error was 'BadDrawable (invalid Pixmap or Window parameter)'.
  (Details: serial 1973 error_code 9 request_code 62 (core protocol) minor_code 0)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the GDK_SYNCHRONIZE environment
   variable to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)
** Message: Initializing gksu extension...
sys:1: PyGIWarning: Caja was imported without specifying a version first. Use gi.require_version('Caja', '2.0') before import to ensure that the right version gets loaded.
```

monsta:
Well, it actually crashes on trying to show icons on the desktop. Moving mouse across the desktop makes some icons appear, flickering, and at that time Caja crashes again.

I'm using GTK+ 3.18.9 here.

libmate-desktop/mate-bg-crossfade.c
XRevan86: Yeah, you're right.

monsta:
Ok, it's a behavior change indeed. With Caja built from https://github.com/mate-desktop/caja/pull/731 there are no crashes.

monsta:
Now testing with GTK+ 3.22.8... interesting, no crashes with this PR but w/o changes in Caja or m-s-d (both are from git master).
Maybe because it's a VM? Just in case, tomorrow I will give it a try on baremetal - my work machine with the same GTK+ (in Debian Testing).

lukefromdc:
The caja changes depend on the mate-desktop changes, but the mate-desktop changes do not depend on the caja changes. The mate-settings-daemon changes are needed by and need the caja changes.

Thus, Monsta's results are consistant with my own testing. In fact, this PR could be used as a standaline fix for the background-fade issue if for any reason it had to be. It can me merged first without creating issues, I ran it for many days that way in my own tests

monsta:
Yes, but something goes wrong with GTK+ 3.18.9...

monsta:
> Now testing with GTK+ 3.22.8... interesting, no crashes with this PR but w/o changes in Caja or m-s-d (both are from git master).
Maybe because it's a VM? Just in case, tomorrow I will give it a try on baremetal - my work machine with the same GTK+ (in Debian Testing).

Ok, it's the same on baremetal.

monsta:
I also have a VM with GTK+ 3.14.5, and the crash mentioned above is reproducible there as well under the same conditions.

raveit65:
All PRs belongs together.
I see no crashes if i updated all 3 packages.
And you need to update first mate-desktop before you build caja and m-s-d against new mate-desktop.

monsta:
I'm just testing possible partial upgrade situations.

raveit65:
> I'm just testing possible partial upgrade situations.

Than you should test 1.16 branch ;-)
1.17x is a development release , if someone update from 1.16 to 1.18 this shouldn't be a problem.
We should mention in release notes that a session restart is required.

For 1.16 this can be a problem, but we need to fix broken fedora 25 and ubuntu-16-10 background handling.
Here the distro maintainers are responsible to update all 3 packages in one rush.
Beside that compiz-reloaded isn't usable in f25 since november last year, backgrounds for dirs are broken in ubuntu too.
So, this is a bugfix for me.
Anyway, i am brave enough to fix f25 from 1.16 branch in case you don't wan't to use it for 1.16.

XRevan86:
That's a lot of reporting :-). I'll poke it some more in a few days.

monsta:
Sure, it's a good bugfix for 1.16. I will test changes for 1.16 later. I think the same issues will appear there, so it doesn't hurt to report them early. :slightly_smiling_face:

monsta:
Good, crashes are gone. Tested with GTK+ 3.14 and 3.18, with Caja from https://github.com/mate-desktop/caja/commit/14b802e267e66f99db00fec1d2e6d262e5b7fb5a or https://github.com/mate-desktop/caja/commit/2a3d4cbef005553cff444aad4caa11f73484e1ea, it doesn't matter.

monsta:
Accidentally, latest changes also fix all-black font on desktop, which I saw with GTK+ 3.14. It was unreadable on dark parts of background. Now it has the proper white color with black shadow (works even if Caja is build from master).

monsta:
Or maybe it was some earlier commit... now I recall that when I saw the white desktop background (before it was fixed), font color was already white with black shadow. Maybe I was confused after testing all these changes in various environments... :slightly_smiling_face:

monsta:
@lukefromdc: can you please give some more testing for the latest changes here and in Caja? I think it should work fine with 3.22, but just in case...

lukefromdc:
This seems to work fine with GTK 3.22. When running without compositing, there is a minor graphical glitch when dragging an icon on the desktop, in which a small square with that portion of the background image momentarily "ghosts" back to the original position. With compositing this does not exist. There have always been a few problems like that with GTK 3.22 and no compositing, though they have varied in exact appearance.

The attached zip file contains a .ogv video showing the icon drag glitch with compositing turned off and GTK 3.22. Again, this is only when compositing is not used.
[Icon_Dragging.ogv.zip](https://github.com/mate-desktop/mate-desktop/files/836810/Icon_Dragging.ogv.zip)

monsta:
Confirmed, I see that glitch even in a VM. Wonder if it was caused by the latest changes? I didn't pay attention before...

raveit65:
I just tested 1.16 branches and i got same results than with 1.17.
with f25 (gtk+-3.22), f24 (mate gtk2) and rhel (mate gtk2).

monsta:
Update: I see the mentioned glitch with compositor as well, on baremetal with GTK+ 3.22.8.
@XRevan86 mentioned that on irc too.

PR #251, opened 2017-02-10
colorsel: port to GtkGrid

raveit65:


raveit65:
@lukefromdc 
Oh sorry,i forgot to mention the gui for it. It's
/usr/bin/mate-color-select

lukefromdc:
Works very nicely, beautiful GUI in my theme and much to be preferred to the native GTK3 version that now comes up in so many places

monsta:
The new GTK+3 version is showing up in many places because we once tried to reduce dependencies on libmate-desktop in various projects. I think mate-color-select is the only tool that uses the old dialog now. I wonder what we should do about all that in the future.

PR #263, opened 2017-05-05
Lose the color-picker and always use hardcoded crosshair cursor

nokangaroo:
Not all cursor themes have a color-picker cursor. This commit uses a hardcoded crosshair cursor for mate-colorsel.c. The existing fallback cursor looks broken


raveit65:
whow, first commit changes about 300 files.
https://github.com/mate-desktop/mate-desktop/pull/263/commits/44241856246c0064b893259a034afc2d2b3c00c0

monsta:
I don't think it's a good idea to commit all the generated files, including .o files :slightly_smiling_face:

nokangaroo:
Where did the .o files come from? My local branch seems to be clean.

monsta:
> I ran git clean -f 

According to git manual, it only cleans files that aren't under version control, so it's too late to run that command now :slightly_smiling_face: 

I'd just reset the branch to two commits earlier (current git master here), then re-apply the fixes and force-push the branch.

PR #267, opened 2017-05-20
Restore Polish translation credits from GNOME

piotrdrag:
This commit restores proper credits and copyrights for the Polish translation that were somehow lost during forking, using historical information from git.gnome.org.


raveit65:
Thank you

PR #277, opened 2017-07-25
Correct little imprecision in man mate-about

cclaw:
The command mate-about doesn't recognize the option --mate-version, it works with --version


raveit65:
Thank you.

PR #281, opened 2017-08-10
gtk show uri deprecations

raveit65:
Please test.
The change in mate-aboutdialog.c can't be tested any more as all MATE applications use gtk_about now.
We need to remove that lib, before someone other starts to fix more deprecations here ;)
@monsta thoughts ?

Honestly, i have no idea where/for what mate-desktop-item.c is used and how we can test this uri change. Any ways, using *NULL* here should be save.

### DIFF
--- a/libmate-desktop/mate-aboutdialog.c
+++ b/libmate-desktop/mate-aboutdialog.c
@@ -162,12 +162,18 @@ default_url_hook (MateAboutDialog *about,
                   const gchar    *uri,
                   gpointer        user_data)
 {
+#if GTK_CHECK_VERSION (3, 22, 0)
+  GError *error = NULL;
+
+  if (!gtk_show_uri_on_window (GTK_WINDOW (about), uri, gtk_get_current_event_time (), &error))
+#else
   GdkScreen *screen;
   GError *error = NULL;
 
   screen = gtk_widget_get_screen (GTK_WIDGET (about));
 
   if (!gtk_show_uri (screen, uri, gtk_get_current_event_time (), &error))
+#endif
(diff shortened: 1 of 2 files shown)


monsta:
Last year I started removing that dialog and switched the last remaining app (mate-about) to GtkAboutDialog, but I didn't like the small height of it. There's a lot of credits in mate-about, so you'd need to scroll a lot to see all of them. You can check it in `cleanup` branch. I'm not sure how to solve this problem.

raveit65:
So want to use mate-aboutdialog.c again?
....not that we have enough other issues/work.

monsta:
I don't want to keep it. Why? I wouldn't work on porting in my branch if I wanted to keep it.
I just don't like the small height of GtkAboutDialog. In most apps there aren't many credits, so you don't have to scroll a lot. But in mate-about there's a lot of them.

raveit65:
Ok, than it's save to drop the lib....

monsta:
It's actually still used by mate-about, the last app to use MateAboutDialog which had been forked from GtkAboutDialog long ago. The cause of the fork was bad behaviour of GtkAboutDialog in GTK+ 3.12. They forced headerbars on all their dialogs in that release. In 3.14 they reverted that, keeping headerbars only when running in GNOME.

raveit65:
I suggest to merge that branch, who cares about credits ?
For the few people who have the time to read credits, it's reasonable to scroll.....

monsta:
Let's test it first, it won't hurt: https://github.com/mate-desktop/mate-desktop/pull/282

Issue #296, opened 2018-01-05
org.mate.interface can-change-accels gschema key doesn't work


tarakbumba:
#### Expected behaviour
Accels can be modified when org.mate.interface can-change-accels is set true.

#### Actual behaviour
After migrating to GTK+3 enabling can-change-accels key doesn't work and a user can not change an accel mouse over.

#### Steps to reproduce the behaviour
1. Set org.mate.interface can-change-accels to true
2. Open a mate component like caja, mouse over on a setting say, Appearance > Show Hidden File option and try to change default value.

#### MATE general version
mate-1.18.0 and up

#### Package version
mate-desktop-1.19.2

#### Linux Distribution
Mageia

monsta:
It won't work with GTK+3, see [here](https://developer.gnome.org/gtk3/stable/GtkSettings.html#GtkSettings--gtk-can-change-accels). We should just remove this option.

tarakbumba:
Yes, you're right. This option should be dropped then. It was nice to configure accels tough.

monsta:
It was nice, though it sometimes could lead to some confusion 🙂 

https://ubuntu-mate.community/t/in-caja-pressing-y-key-opens-new-tab-cant-rename-files/13551

akshaypuradkar:
so how do i change the accels?

PR #297, opened 2018-01-05
drop non-working can-change-accels option from GSettings

monsta:
it doesn't work with GTK+3, see https://github.com/mate-desktop/atril/issues/247, https://github.com/mate-desktop/mate-calc/issues/30, https://github.com/mate-desktop/mate-desktop/issues/296

this PR bumps version so that m-s-d can rely on it

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,5 @@
+### mate-desktop 1.19.3
+
 ### mate-desktop 1.19.2
 
   * Translations update
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,6 @@
 m4_define([mate_platform], [1])
 m4_define([mate_minor], [19])
-m4_define([mate_micro], [2])
+m4_define([mate_micro], [3])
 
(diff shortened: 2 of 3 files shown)


raveit65:
....for better testing with a new tarball ;-)

monsta:
Oops, I didn't think about the NEWS 🙂

lukefromdc:
NEWS commit is in, just doesn't have anything added to it other than the version bump yet

raveit65:
> NEWS commit is in, just doesn't have anything added to it other than the version bump yet

This is OK, so we have one step left for doing the release commit and tagging it.

lukefromdc:
Also, built packages may have to specify that they break older versions of mate-settings-daemon (they do-missing schema), but as mate-settings-daemon is not a build or runtime dependency of mate-desktop, the only thing needed here may be a note in NEWS.

monsta:
Yes, it's on my list to do for https://github.com/mate-desktop/debian-packages

PR #301, opened 2018-01-26
building: remove < gtk+-3.22 bits and bump required gtk+ version

raveit65:


raveit65:
Btw. i used 'grep -nr GTK_CHECK_VERSION * ' to find old gtk+ bits, should be sufficient.

PR #310, opened 2018-04-10
Support background fallback on HiDPI

vkareh:
This change is a super hacky way of supporting the fallback desktop background (without caja) on HiDPI displays. It does this by doing two things: providing a function for scaled-up background surfaces using cairo_scale; and triggering a change signal when the entire background schema is re-loaded, and so triggering a redraw.

### DIFF
--- a/libmate-desktop/mate-bg.c
+++ b/libmate-desktop/mate-bg.c
@@ -276,6 +276,9 @@ mate_bg_load_from_preferences (MateBG *bg)
 
 	mate_bg_load_from_gsettings (bg, settings);
 	g_object_unref (settings);
+
+	/* Queue change to force background redraw */
+	queue_changed (bg);
 }
 
 /* This function loads default system settings */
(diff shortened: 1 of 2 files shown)


monsta:
Maybe calculate the scale inside the existing function here, not in m-s-d, to avoid API change?

vkareh:
@monsta - I tried that, but then caja would draw a doubly-large background. Caja uses this same function and it renders fine. I hacked at both caja and m-s-d to see what the difference was in how they draw the background and even paring it down to a super basic version of it, with the same code on both sides, m-s-d background was always half the size of caja's...

monsta:
Ah right, I didn't think about it. But then Caja could be made to not scale the background at all, to avoid double scaling?

raveit65:
Guys, i like to merge it.
Looks like this is ready to to go, but there is a open question.
Could team members please communicate?

flexiondotorg:
I've added this patch to Debian and Ubuntu.

vkareh:
Yeah, I'm still not satisfied with this approach, but it seems to work well enough for now... I've looked at this issue quite a bit and haven't been able to solve it any other way, unfortunately. Maybe someone else can try to tackle the issue from a different angle?

monsta:
Could you find where Caja does its own scaling?

vkareh:
@monsta - yes, I found where that happens, but even passing the same hard-coded values from caja and m-s-d to the mate-desktop, the caja background was still twice as large as the m-s-d one... very confusing (which is why I made this hack) :confused:

monsta:
Weird. Well, then we have no other choice, at least for now.

monsta:
Ok, now a new release is needed, so that m-s-d would require it for build

raveit65:
I would wait with a release to catch up further changes..... Fedora and ubuntu is already updated

PR #313, opened 2018-04-25
Change default cursor-size to 24px

vkareh:
The MATE cursor theme does not have an 18px cursor size, so having
a 18px as the default setting does not make sense. Using 24px looks
better in modern displays and allows for proper HiDPI scaling.

Relevant: mate-desktop/mate-control-center/pull/342


lukefromdc:
This one I have no way to test due to some long forgotten configuration issue preventing me from changing my cursor off of my (long ago chosen) system default, will have to leave to the rest of the team
https://github.com/mate-desktop/mate-control-center/pull/342#issuecomment-384157623

raveit65:
Ok, merging this as related m-c-c commit is already merged.

PR #320, opened 2018-05-26
Add toggle for synchronizing QT_SCALE_FACTOR to GDK_SCALE_FACTOR

vkareh:
When enabled (default), m-s-d will attempt to set QT_SCALE_FACTOR during the init phase.
When disabled, the user can then choose to control this env var elsewhere.

* Requires mate-desktop/mate-settings-daemon/pull/225
* Fixes mate-desktop/mate-session-manager/issues/158

### DIFF
--- a/schemas/org.mate.interface.gschema.xml.in
+++ b/schemas/org.mate.interface.gschema.xml.in
@@ -174,7 +174,12 @@
       <default>0</default>
       <range min="0" max="2"/>
       <summary>Window Scaling Factor</summary>
-      <description>Internal scale factor that maps from window coordinates to the actual device pixels. On traditional systems this is 1, but on very high density displays (e.g. HiDPI, Retina) this can be a higher value (often 2). Set to 0 to auto-detect.</description>
+      <description>This controls the GTK scale factor that maps from window coordinates to the actual device pixels. On traditional systems this is 1, but on very high density displays (e.g. HiDPI, Retina) this can be a higher value (often 2). Set to 0 to auto-detect.</description>
+    </key>
+    <key name="window-scaling-factor-qt-sync" type="b">
+      <default>true</default>
+      <summary>Scaling Factor for QT appllications</summary>
+      <description>This setting determines whether MATE controls the scale factor for QT applications. Enable to synchronize with the GTK scale factor when initializing the session, disable to control this value elsewhere. Requires restarting your session.</description>
     </key>
   </schema>
 </schemalist>


assen-totin:
Guys, a quick question since I observe the following phenomenon that is related to this commit: 
- My laptop's display is Retina, so when I use it, GDK scaling and QT scaling are both set to 2 my m-s-d. Works fine.
- 99% of the time I only use an external display (laptop lid closed, its display is off), which is 1920x1080 and only 100 dpi. On this display m-s-d keeps the GDK scaling at 1 (as expected), but still sets QT scaling at 2, making QT apps look quite huge. 
- I currently have this package installed: mate-settings-daemon-1.20.2-1.fc28.x86_64 - is this commit included in this version of the package (which would mean I need to look at the m-s-d code further)? Or should I wait for this commit to appear in some next release?

vkareh:
@assen-totin - the issue here is that the QT setting gets applied during session start, and MATE is unable to change it after the fact (it has to do with how Dbus allows to set env variables during the init phase only). Since your primary monitor is the laptop screen (as seen by X11), the setting is based on that when you log in, regardless of your external monitor's resolution.

This is one instance in which you want this setting here.

@raveit65 - is this backported to 1.20 on f28?

raveit65:
yes, available since march.
 mate-desktop-1.20.2 does have this commit
https://github.com/mate-desktop/mate-desktop/commits/1.20
https://koji.fedoraproject.org/koji/buildinfo?buildID=1088353
Same for m-s-d-1.20.2
https://github.com/mate-desktop/mate-settings-daemon/commits/1.20
https://koji.fedoraproject.org/koji/buildinfo?buildID=1088384

vkareh:
@assen-totin - do you have this setting in dconf? See if it's there and try toggling it to see if it changes anything for you (don't forget to restart your session after changing the setting)

assen-totin:
Thanks for the clarifications - yes, the setting is there and has a value of 0, which seems to mean "auto". It seems fine to me as I want to keep the scaling when the Retina display is in use and not have it when using the external display. 

To clarify the workflow a bit, I'm on F28 and I have to log on using the laptop display as the login window is not shown on the external display; maybe this is the moment m-s-d decides to set scaling to 2 for QT and never reconsiders even after the only display left has lower DPI. After I log in, I guess, my config kicks in and the laptop display gets disabled, leaving the external display the only one. 

Up to F25, I think, I did get the GUI login prompt on my external display and so I never had to open the laptop's lid - but since F26 something changed and the login window is only shown on the laptop's screen (guess the default for LightDM used to be to "mirror" displays, but is now not). However, Plymouth does mirror the laptop's screen during boot, so I get my LUKS prompt on the external display. So, maybe there is away to revert LightDM to its previous behaviour and that would cure the problem for me?

vkareh:
I think F28 changed from LightDM to Slick Greeter (@raveit65 can confirm that), so your issue might be there.

The setting you probably want to look at is actually `window-scaling-factor-qt-sync`, which only takes True/False. What it does is allows you to have your retina display scale automatically as it usually does, but ignore the QT_* variables. If you set it to `false`, you'll most likely invert the behavior: QT apps will be small when on the Retina display, and normal when on the external monitor. If you use the external monitor 99% of the time, you probably should turn this flag to `false`.

raveit65:
@assen-totin 
We switched to using slick-greeter for f28 as login-manager, another frontend for lightdm which supports HIDPI.
For xrandr config with HIDPI and non-HIDPI monitors take a look at https://wiki.archlinux.org/index.php/HiDPI#Multiple_displays
Btw. The icon from your mate-applet-streamer looks a bit blurred with HIDPI ;) maybe you like to do something like this  https://github.com/mate-desktop/mate-panel/commit/5c4eb86ea1bb117805cc8fccb4ba3d96a53f2813

raveit65:
Btw. I see slick-greeter login on boths, notebook monitor and external monitor with my old msi fx700 notebook, of course without HIDPI resolutions.

Issue #319, opened 2018-05-24
Need more useful API


yetist:
Can the new API in gnome-desktop be implemented in mate-desktop? For example, like the following:
```
char *        gnome_get_language_from_locale    (const char *locale,
                                                 const char *translation);
char *        gnome_get_country_from_locale     (const char *locale,
                                                 const char *translation);
char **       gnome_get_all_locales             (void);
gboolean      gnome_parse_locale                (const char *locale,
                                                 char      **language_codep,
                                                 char      **country_codep,
                                                 char      **codesetp,
                                                 char      **modifierp);
char *        gnome_normalize_locale            (const char *locale);
gboolean      gnome_language_has_translations   (const char *code);
char *        gnome_get_language_from_code      (const char *code,
                                                 const char *translation);
char *        gnome_get_country_from_code       (const char *code,
                                                 const char *translation);
gboolean      gnome_get_input_source_from_locale (const char  *locale,
                                                  const char **type,
                                                  const char **id);
```

raveit65:
Why not.......... for what do you needed?
Looks like an tool for language and localization.

yetist:
Want to write a user-manage tool for mate-control-center, it need some api for user language choose.

raveit65:
Great idea, a tool for switching the language is really missing since i switch fedora to use slick-greater instead of lightdm-gtk.
Lightdm-gtk offers an option to switch session language but slickgreeter doesn't.
So i agree with adding those API.
@monsta @lukefromdc @sc0w @vkareh 
Thoughts?

lukefromdc:
I don't know much about working with languages (I only speak one and purge unused locales), but this strikes me as a missing part of slick-greeter if DE's now expect it in the session greeter. If nothing else, it could make the greeter useless on a machine used by two different people who don't speak any common language between them

No problem with adding this to MATE to deal with the reality of possible randomly sourced greeters, but I am not in a good position to test it.

raveit65:
A language tool is also useful for posting screenshots/logs here in one language which everybody understand ;)

flexiondotorg:
I think improving the options for switching session locales is good :smiley: 

What I'd like to see is a fork of mate-desktop with the required API changes implemented in a branch and the proposed tool for switching locales in another repo so we can better review both. 

@yetist  Are you going to keep the `gnome_` prefixes on function names or will you be changing them to `mate_`?

monsta:
[...]

PR #325, opened 2018-06-16
add mate-language apis

yetist:
Add mate language/locale api from gnome-desktop, see #319.
Please review, thank you.


yetist:
Here is a simple test code for these new apis:
```
$ cat test.c 
/* gcc test.c `pkg-config --cflags --libs mate-desktop-2.0` */
#include <stdio.h>
#include <glib.h>
#define MATE_DESKTOP_USE_UNSTABLE_API
#include <libmate-desktop/mate-languages.h>

void test_one_locale(const gchar *locale)
{
	char *lang, *country, *norm_locale;
	char *language_codep, *country_codep, *codesetp, *modifierp;

	lang = mate_get_language_from_locale (locale, NULL);
	country = mate_get_country_from_locale (locale, NULL);
	norm_locale = mate_normalize_locale (locale);

	g_print("%s:\nlang=%s, country=%s, locale=%s\n", locale, lang, country, norm_locale);


	if (mate_parse_locale (locale, &language_codep, &country_codep, &codesetp, &modifierp)) {
		g_print("lang_codep=%s, country_code=%s, code=%s, modifier=%s\n",
				language_codep,
				country_codep,
				codesetp,
				modifierp);
		if (mate_language_has_translations(language_codep)) {
			g_print("Has translation\n");
		} else {
			g_print("No translation\n");
		}
	}
	putchar('\n');

	g_free(lang);
	g_free(country);
	g_free(norm_locale);

	g_free(language_codep);
	g_free(country_codep);
	g_free(codesetp);
	g_free(modifierp);
}

void test_locales(void)
{
	char **all;
	guint i, len;

	all= mate_get_all_locales ();
	len = g_strv_length (all);

	for (i =0; i < len; i++) {
		test_one_locale(all[i]);
	}
	g_strfreev(all);
}

int main(int argc, char **argv)
{
	if (argc == 2) {
		test_one_locale(argv[1]);
	} else {
		test_locales();
	}
	return 0;
}

```

raveit65:
But you should see your USA locale with the poc-tester?
Or does Americans live without any locale ?  :)

lukefromdc:
I have zero experience working with languages. Two locales are kept here and all others purged, though you can in fact run without any and default to "C" as it is called for some reason. That happens if locales are broken outright. An English speaker will notice warnings only, anyone else gets a broken system when that happens.  Beyond that, I have put zero study or time into language/locale matters other than saving the 500MB of space from keeping only what I use.

Anyway, I don't know enough about this to test it properly. I do not even know what poc-tester is.

raveit65:
It's OK, for me the poc-tester works fine, but are you really sure you don't use  `en_US`  in the states ? 
```
en_US.UTF-8:
lang=English (United States), country=United States (English) [UTF-8], locale=en_US.UTF-8
lang_codep=en, country_code=US, code=UTF-8, modifier=(null)
Has translation
```

raveit65:
> I do not even know what poc-tester is.

poc=proof of concept
tester = test.c from second post

Any way, `test.c` works fine, no need to test it again. Only a code review is needed by someone which have more knowledge than me.

monsta:
Should we add this code to this library if it would be only used by one tool?

raveit65:
Updated poc-tester works fine.
```
Current locale: de_DE.UTF-8
[_locale]:		lang=Deutsch (Deutschland), country=Deutschland (Deutsch), locale=de_DE.UTF-8
[_parse_locale]:	lang_code=de, country_code=DE, code=UTF-8, modifier=(null)
[_code]:		lang=Deutsch, country=Deutschland, Has translation

Current locale: uk_UA.UTF-8
[_locale]:		lang=Українська, country=Україна (Українська), locale=uk_UA.UTF-8
[_parse_locale]:	lang_code=uk, country_code=UA, code=UTF-8, modifier=(null)
[_code]:		lang=Українська, country=Україна, Has translation
```
etc.

raveit65:
Thanks

PR #332, opened 2018-08-24
interface: add key to enable/disable primary paste from middlemouse

lukefromdc:
Needed for https://github.com/mate-desktop/mate-settings-daemon/pull/234   Default is "enabled" to avoid any change in expected behavior with a default install.

### DIFF
--- a/schemas/org.mate.interface.gschema.xml.in
+++ b/schemas/org.mate.interface.gschema.xml.in
@@ -181,5 +181,12 @@
       <summary>Scaling Factor for QT appllications</summary>
       <description>This setting determines whether MATE controls the scale factor for QT applications. Enable to synchronize with the GTK scale factor when initializing the session, disable to control this value elsewhere. Requires restarting your session.</description>
     </key>
+    <key name="gtk-enable-primary-paste" type="b">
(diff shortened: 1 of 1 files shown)


raveit65:
Guys, i think we need to bump required version of mate-desktop in m-s-d and m-c-c.
Not tested, but I am pretty sure m-s-d will crash because of a missing gsettings schema if an older version of mate-desktop is installed.

lukefromdc:
I think mate-settings-daemon already requires 1.21, meaning we will need
a minor version bump for all three. At release it will be a requirement for 
mate-desktop 1.22

raveit65:
Ok, i miss that m-s-d is already requires mate-desktop-1.21.0
https://github.com/mate-desktop/mate-settings-daemon/blob/master/configure.ac#L57
Than we only need to bump requires in m-s-d and m-c-c to mate-desktop-1.21.1.
I will do a pre-release bump for mate-desktop to 1-21.1 and update the the requirement for the other packages.
This should be sufficient, a new m-d release isn't needed for the moment.

raveit65:
Done.

PR #335, opened 2018-09-26
Add new setting to control XKB modifier latching

cwendling:
Add new setting *org.mate.accessibility-keyboard.stickykeys-latch-to-lock* to control whether latching is enabled together with sticky keys.

Default value is for compatibility with current behavior, although I'd rather see that disabled by default as it is fairly confusing for the newcomer.

See https://github.com/mate-desktop/mate-settings-daemon/pull/237 and https://github.com/mate-desktop/mate-control-center/pull/374.


lukefromdc:
I have no idea what latching is, but installing this does not break current m-s-d from master. About to test the other two PR's

PR #338, opened 2018-11-09
avoid redefined warning of MATE_DESKTOP_USE_UNSTABLE_API

yetist:
It show redefined warning  when build use meson, so change the code.


yetist:
OK, close this pr.

PR #346, opened 2019-01-03
update copyright  to 2019

yetist:
I think it's time to update copyright  to 2019.


raveit65:
Thanks. Yes, it's time for update Copyright.
I think you can commit directly without a PR for such easy changes.

PR #358, opened 2019-02-05
mate-colorsel: avoid deprecated 'gtk_widget_modify_bg'

sc0w:
The expected: the same colors here:

### DIFF
--- a/libmate-desktop/mate-colorsel.c
+++ b/libmate-desktop/mate-colorsel.c
@@ -1226,6 +1226,26 @@ palette_change_color (GtkWidget         *drawing_area,
   g_free (current_colors);
 }
 
+static void
+override_background_color (GtkWidget *widget,
+                           GdkRGBA   *rgba)
+{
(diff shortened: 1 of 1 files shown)


raveit65:
Do you really want to fix deprecation here?
In the end all this code needs to be dropped for Gtk+4 because there is no replacement for deprecated GtkHSV.
https://developer.gnome.org/gtk3/stable/GtkHSV.html
Sadly, the UI with the color-cycle doesn't have any future.
Your decision....i only wanted inform you.

sc0w:
@raveit65 ok, I noticed it now

I like to fix deprecations everywhere, I think it is early to thinking in gtk 4.

sc0w:
@lukefromdc  Is not easy to fork GtkStatusIcon, believe me, it has a loooot of mooore deprecated code internally.

sc0w:
I think GtkHSV can be forked to mate

PR #362, opened 2019-02-09
MateHSV implementation (GtkHSV fork)

sc0w:
Note: we have a lot of lines here, please don't merge to the 1.22 release

### DIFF
--- a/docs/reference/mate-desktop/mate-desktop-sections.txt
+++ b/docs/reference/mate-desktop/mate-desktop-sections.txt
@@ -288,3 +288,25 @@ mate_language_has_translations
 mate_desktop_prepend_terminal_to_vector
 </SECTION>
 
+<SECTION>
+<FILE>mate-hsv</FILE>
+<TITLE>MateHSV</TITLE>
+MateHSV
+mate_hsv_new
+mate_hsv_set_color
+mate_hsv_get_color
+mate_hsv_set_metrics
+mate_hsv_get_metrics
+mate_hsv_is_adjusting
+<SUBSECTION Standard>
+MATE_TYPE_HSV
+MATE_HSV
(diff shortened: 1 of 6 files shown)


yetist:
Please rebase on master, all check will be passed.

monsta:
@raveit65: what if we port `mate-color-select` to GTK+3 version of color selection widget (instead of the current color wheel)? It's our only app that uses the old widget.

lukefromdc:
The color wheel but NOT the palette grid is used by caja's "backgrounds and emblems" dialog for adding a new background color for folders (not the desktop). We could port both of those to GTK3's color selection widget as everything else already has been, or at the risk of higher maintainance port all our other color selectors back to the color wheel.

Our version is easier to use and in my opinion just plain a better version, but unless we plan to use it more widely we should probably drop it as it is unused code to most users.

raveit65:
For me the new color selection from gtk+ isn't so bad and i don't think it is more complicated to use than our traditional color cycle.
I like that i can use transparent rgba colors definition.

monsta:
Well, for example, mate-system-monitor already uses the new color selector (for CPU colors and stuff).

sc0w:
@monsta  with the "new color selector" the user can click a color anywhere on the screen to select that color?

monsta:
Good question... I didn't see that there.

raveit65:
Beside that the color-cycle is replaced with something new, our GUI has 2 functions which aren't anymore in the new gtk3 gui.
bildschirmfoto zu 2019-02-13 21-38-57
1. The color picker is gone.
2. You can't type rgb values and convert them to hex values.

That the new gtk3 GUI use rgba colors is one advantage.

So , i think it isn't worse to fork GtkHSV.

PR #367, opened 2019-03-11
mate-hsv: avoid duplicate initialization in variable

sc0w:
```
mate-hsv.c:913:19: warning: Value stored to 'priv' during its initialization is never read
  MateHSVPrivate *priv = hsv->priv;
                  ^~~~   ~~~~~~~~~
```


sc0w:
I only change the commit message in last change

PR #375, opened 2019-03-19
I18n mate about dialog

rbuj:


raveit65:
I just split the update of the resource file for transifex in an extra commit.
This will make it easier to cherry-pick the commits to a stable branch without breaking translations.
And this will make it easier for me to test the commit with a patch in a rpm,.....
.......editing commit is just boring....
Would be nice when you can do that like this in new PRs for yourself, please.

PR #379, opened 2019-04-07
Use mate-desktop icon instead of mate for mate-about

rbuj:


rbuj:
@lukefromdc 

It guarantees that icon lookup is always performed in hicolor folder, as it has a different name to icon theme. Otherwise, the icon provided with mate-desktop is never used if mate theme is selected, since themes have higther priority for icon lookups.

Another opproach is to leave it as before, install mate icon in both places (hicolor and theme folders) using the same icon name: "mate". Thus mate icon lookup is always resolved even if mate icon theme is not installed, or the mate theme is not selected.

PR #393, opened 2019-06-10
schemas: Define schema for a default messenger

rbuj:
Defines a schema that allows users to specify a default instant messaging application.
Default initialized to 'pidgin'.

### DIFF
--- a/po/POTFILES.in
+++ b/po/POTFILES.in
@@ -22,6 +22,7 @@ schemas/org.mate.applications-at-mobility.gschema.xml
 schemas/org.mate.applications-at-visual.gschema.xml
 schemas/org.mate.applications-browser.gschema.xml
 schemas/org.mate.applications-calculator.gschema.xml
+schemas/org.mate.applications-messenger.gschema.xml 
 schemas/org.mate.applications-office.gschema.xml
 schemas/org.mate.applications-terminal.gschema.xml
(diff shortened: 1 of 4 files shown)


vkareh:
This works fine, pending an approval on https://github.com/mate-desktop/mate-control-center/pull/457, but should we default it to just empty, since MATE doesn't specifically endorse any specific IM client?

rbuj:
Similar case with Web Browser, by default is set to 'mozilla'. It should be updated to ['firefox'](https://wiki.debian.org/Firefox).
https://github.com/mate-desktop/mate-desktop/blob/d53d2be25eded44f81c126c9d904d21774f7ee42/schemas/org.mate.applications-browser.gschema.xml#L4

lukefromdc:
I think that will need to default to empty for that reason.

rbuj:
@vkareh @lukefromdc done

raveit65:
I will reword commit body to match latest default.

lukefromdc:
It would seem to me we should now increment version to 1.23.1 since we've added a new schema that a mate-control-center change will depend on

raveit65:
Ok, i will make a new mate-desktop release.

raveit65:
I just bumped version to 1.23.1 https://github.com/mate-desktop/mate-desktop/commit/f3a7cc35331b06c69a8c8c07da372cf73bbc5d13

PR #395, opened 2019-06-15
CI: fix meson distcheck

raveit65:


raveit65:
I noticed that tarballs with created with `ninja -C build dist` doesn't include the the normal auto generated files for building with autotools, like configure and Make files.
I guess this is the expected behaviour.
Well, autogen.sh is in tarball, so this files can be generated easily,
and this is better than to have the unneded .gmo files in tarballs since port to gettext.

PR #396, opened 2019-06-22
using autotools to generate tarballs

raveit65:
This is better because `ninja -C _build dist` do not create auto-generated files


raveit65:
Result is here https://travis-ci.org/mate-desktop/mate-desktop/jobs/549041921#L2652

Issue #399, opened 2019-07-05
General questions about C coding


tamplan:
Hello everyone,

Not really an issue but questions about Mate development in C language and i don't where to post this subject.

I have made some PR accepted by coping/pasting the same piece of code several times and sc0w have made this https://github.com/mate-desktop/mate-polkit/pull/49 whose code could be picked up elsewhere due to Gtk deprecated functions in others repositories whose use too GtkStock to replace.

Is it possible to imagine a new repository called, for example, **Mate Gtk** or something else to regroup several pieces of code about Gtk deprecated functions in a shared library called, for example, **libmategtk** used by various over Mate repositories?

What are the pro and con about this solution in term of size of applications on disk and memory used by the Mate Desktop project?

This should be a good place to regroup code and prepare future depreciation using GTK 4.0 for example.

Thanks by advance for any post about this subject.

Regards

vkareh:
We regularly have conversations about this. One of the issues with this is that deprecated code is still shipped, which means any such effort cannot be used until MATE moves to GTK4, and at that time even the code that provided whatever functionality might no longer be compatible with the rest of GTK. Another issue is that of maintenance, this is a huge undertaking and who will take care of it?

The simple answer is yes, we will fork parts of GTK as needed to provide necessary functionality. We don't yet know when or whom or how, though.

vkareh:
Case in point: mouse scrolling on notebook tabs: [GNOME/gtk@ad48f4d5](https://gitlab.gnome.org/GNOME/gtk/commit/ad48f4d52bbac6139dd829fcc421ad16441f34d2). this is a huge piece of functionality that was removed and there has been a lot of work done to bring it back to MATE.

It's not a simple "clone GtkNotebook into another repo" solution. It requires reimplementing this function in every program that needs it. Otherwise we would need to modify the widget class or provide an interface which will not be compatible with GtkNotebook moving forward, plus wouldn't work with any application using GtkNotebook anyway, so it still needs a per-application code change.

Ideally we can, at the libmate-desktop level, make _all_ GtkNotebook instances listen to the scroll event and call our own clone of `gtk_notebook_scroll`, but 1) how do you do that in a way that works for all GtkNotebooks in the exact same way, taking into account their widget container hierarchy? and 2) sounds like that would break a lot of applications.

[...]

PR #400, opened 2019-07-07
Avoid GtkStock

tamplan:
Add new function to be shared by several modules/applications

### DIFF
--- a/libmate-desktop/mate-gtkstock-replacement.c
+++ b/libmate-desktop/mate-gtkstock-replacement.c
@@ -1,0 +1,60 @@
+/* -*- Mode: C; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- */
+/* mate-gtkstock-replacement.c - MATE Desktop button management for dialog
+
+   Copyright (C) 2019, ZenWalker
+   All rights reserved.
+
+   This file is part of the Mate Library.
+
+   The Mate Library is free software; you can redistribute it and/or
+   modify it under the terms of the GNU Library General Public License as
+   published by the Free Software Foundation; either version 2 of the
+   License, or (at your option) any later version.
+
+   The Mate Library is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+   Library General Public License for more details.
+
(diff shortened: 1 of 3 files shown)


tamplan:
Seen check failed, i search to resolve this...

tamplan:
I forgot to add to track two new files, this is done now but the compile process does not start now...

sc0w:
@mate-desktop/core-team calling to others...

I think it can be on `mate-desktop-utils.c` for one more function

libmate-desktop/mate-gtkstock-replacement.c
```diff
@@ -0,0 +1,60 @@
...
+                        const gchar *button_text,
+                        const gchar *icon_name,
+                        gint   response_id)
+{
+    GtkWidget *button;
+
+    button = gtk_button_new_with_mnemonic (button_text);
+
+    // Set button image only if icon_name is set
+    if (icon_name != NULL) {
```
sc0w: seems you modified the original function with this change

we need this function only to work with icons, if we don't like icons in buttons we have already `gtk_dialog_add_button` , so, we don't need this condition

tamplan: Yes, i modify the function to test if icon is need or not.
The function call some others functions of GTK in one line:
`    gtk_button_set_use_underline (GTK_BUTTON (button), TRUE);
    gtk_style_context_add_class (gtk_widget_get_style_context (button), "text-button");
    gtk_widget_set_can_default (button, TRUE);
    gtk_widget_show (button);
    gtk_dialog_add_action_widget (GTK_DIALOG (dialog), button, response_id);
`

tamplan: i have copy/paste header of **mate-gtkstock-replacement.c** from another file, update and giving you credit of this function, feel free to comment here too to advice me.

tamplan:
> @mate-desktop/core-team calling to others...
> 
> I think it can be on `mate-desktop-utils.c` for one more function

Please read here https://github.com/mate-desktop/mate-polkit/pull/52/files, you can see directly than the new file is in charge about only one subject.

Like indicating here https://github.com/mate-desktop/mate-desktop/issues/399#issuecomment-508993566, I'm novice in C programming but I had understood that several files c were finally compiled into one (libmate-desktop), including the correct h file declare the function called from the library.

tamplan:
Maybe planned in future release...

PR #407, opened 2019-08-02
schema: Move mouse schemas to mate-settings-daemon

vkareh:
Multi-part patch:
* https://github.com/mate-desktop/mate-desktop/pull/407
* https://github.com/mate-desktop/mate-settings-daemon/pull/293
* https://github.com/mate-desktop/mate-control-center/pull/494


raveit65:
@vkareh 
I think it's save to pre-bump version in configure and NEWS here in PR.

schemas/org.mate.peripherals-mouse.gschema.xml
```diff
@@ -15,6 +19,11 @@
       <summary>Motion Threshold</summary>
       <description>Distance in pixels the pointer must move before accelerated mouse motion is activated.  A value of -1 is the system default.</description>
     </key>
+    <key name="accel-profile" enum="org.mate.peripherals-mouse.AccelProfile">
+      <default>'adaptive'</default>
+      <summary>Acceleration profile</summary>
+      <description>Acceleration profile for connected mice. The acceleration profile can be either 'adaptive', which adapts the acceleration depending on the mouse movement, or 'flat, which accelerates by a device-specific constant factor derived from the configured pointer speed.</description>
```
monsta: Small thing: missing single quote after "flat" in description.

vkareh:
@raveit65, @monsta - I've moved the touchpad gschema to this repo. Please review

raveit65:
@vkareh 
Hmm, why not using m-s-d for both schemas?
The schemes are only used by m-s-d and m-c-c, and we need only to bump version in msd + updating `MSD_REQUIRED=` in m-c-c.
With you solution we need to this for 3 packages.

Well, i am not against your solution if there is another serious reason which i missed :)

vkareh:
Huh, that's a good point. I can move all the peripheral schemas to m-s-d

monsta:
Yes, it makes more sense to move mouse settings to m-s-d instead. I would do it in a separate commit/PR though...

I think `org.mate.peripherals-keyboard` could be moved too.

schemas/org.mate.peripherals-mouse.gschema.xml
```diff
@@ -15,6 +20,11 @@
       <summary>Motion Threshold</summary>
       <description>Distance in pixels the pointer must move before accelerated mouse motion is activated.  A value of -1 is the system default.</description>
     </key>
+    <key name="accel-profile" enum="org.mate.peripherals-mouse.AccelProfile">
+      <default>'default'</default>
+      <summary>Acceleration profile</summary>
+      <description>Acceleration profile for connected mice. The acceleration profile can be either 'adaptive', which adapts the acceleration depending on the mouse movement, or 'flat', which accelerates by a device-specific constant factor derived from the configured pointer speed. If a mouse doesn't support the configured profile, 'default' will be used.</description>
```
monsta: We can mention the default at the beginning, like GNOME does:

`
<description>Acceleration profile used for connected mice. The acceleration profile can be set to either default ('default') which uses the default acceleration profile for each device, flat ('flat'), which accelerates by a device specific constant factor derived from the configured pointer speed, or adaptive ('adaptive') which adapts the acceleration depending on the mouse movement. If a mouse doesn't support the configured profile, 'default' will be used.</description>`

monsta:
Side note: current Travis build of https://github.com/mate-desktop/mate-settings-daemon/pull/293 failed because the moved schema is still listed in `po/POTFILES.in` file. It should be addressed when moving the mouse schema to m-s-d.

vkareh:
Okay, how about now @raveit65 ?

raveit65:
LGTM.
Ok, lets start with merging this one.
Than i can make a new release for fixing travis builds of the other PRs.

lukefromdc:
The version bump for this left out meson.build, so meson builds of 1.23.2 break building mate-settings-daemon 1.23.1 and show the wrong version in mate-about.
https://github.com/mate-desktop/mate-desktop/pull/409 fixes this

PR #408, opened 2019-08-07
Replace deprecated gtk css provider get default

raveit65:
Please test.

### DIFF
--- a/libmate-desktop/mate-bg-crossfade.c
+++ b/libmate-desktop/mate-bg-crossfade.c
@@ -267,11 +267,16 @@ tile_surface (cairo_surface_t *surface,
     }
     else
     {
+        static GtkCssProvider *provider = NULL;
         GtkStyleContext *context;
         GdkRGBA bg;
+
+        if (provider == NULL)
+              provider = gtk_css_provider_new ();
+
         context = gtk_style_context_new ();
         gtk_style_context_add_provider (context,
-                                        GTK_STYLE_PROVIDER (gtk_css_provider_get_default ()),
+                                        GTK_STYLE_PROVIDER (provider),
                                         GTK_STYLE_PROVIDER_PRIORITY_THEME);
         gtk_style_context_get_background_color (context, GTK_STATE_FLAG_NORMAL, &bg);
         gdk_cairo_set_source_rgba(cr, &bg);
--- a/libmate-desktop/mate-bg-crossfade.c
+++ b/libmate-desktop/mate-bg-crossfade.c
@@ -40,28 +40,28 @@
 
 struct _MateBGCrossfadePrivate
 {
-	GdkWindow       *window;
-	GtkWidget       *widget;
-	int              width;
-	int              height;
-	cairo_surface_t *fading_surface;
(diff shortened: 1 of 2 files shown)


raveit65:
I just updated PR, but i am not sure if this is better.

raveit65:
PR is updated again, please review.

raveit65:
`if (surface != NULL) `
But this is old code, i didn't change that.

lukefromdc:
Question: when tile_surface is rerun (which could happen a lot), can it detect the previous cssprovider from the last time it ran? I suspect but am not sure that it cannot. If not we still get a new cssprovider everytime (and leak the old one)-and probably got it with the old gtk_css_provider_get_default as well

lukefromdc:
We had a similar problem in Pluma with the override-font code, and solved it by using both a static cssprovider and by using a static boolean to determine if it has already been added and this worked. Commit was d3757dfa7b822dba1dc1b686ae60fdf77a097cf2, code change was

```
@@ -686,7 +686,7 @@ pluma_override_font (const gchar          *item,
 		     PangoFontDescription *font)
 {
 	static gboolean provider_added = FALSE;
-	GtkCssProvider *provider;
+	static GtkCssProvider *provider; /*We need to keep this as long as Pluma is running*/
 	gchar          *prov_str;
 	gchar          *css;
 	gchar          *family;
@@ -709,7 +709,8 @@ pluma_override_font (const gchar          *item,
 				pango_font_description_get_size (font) / PANGO_SCALE,
 				pango_font_description_get_size_is_absolute (font) ? "px" : "pt");
 
-	provider = gtk_css_provider_get_default ();
+	if (!provider_added)
+		provider = gtk_css_provider_new ();
 
 	prov_str = gtk_css_provider_to_string (provider);
 
@@ -729,7 +730,7 @@ pluma_override_font (const gchar          *item,
 
 	if (!provider_added)
 	{
-		GSettings *settings;
+		static GSettings *settings; /*We need this for the life of the provider and program*/
 		settings = g_settings_new ("org.mate.interface");
 		g_signal_connect (settings,
 				  "changed::" "font-name",

```

raveit65:
Should i close PR?
I don't get your mentioned problem.

sc0w:
@lukefromdc seems you don't know how it works the static variables

in the line

`static GtkCssProvider *provider = NULL;`

`provider = NULL` works only one time, the next times this line is ignored and provider have the previous value.

```
        if (provider == NULL)
              provider = gtk_css_provider_new ();
```

this condition only works the first time, when `provider` is `NULL`, the next times `gtk_css_provider_new` is never called because `provider` already exists

@lukefromdc the pluma code you mentioned, probably can be improved, I will investigate it, I think the **static** gboolean variable isn't needed.

lukefromdc:
OK, https://github.com/mate-desktop/pluma/pull/495 works fine and oh yeah: I now see how this works: since the static variable is kept the NULL test should work same as if it was a global or part of the toplevel structure.  The red herring which confused me was a report of a build error in the mate-screensaver PR, https://github.com/mate-desktop/mate-screensaver/pull/203#issuecomment-519559470

PR #415, opened 2019-10-13
build: introduce RELEASE_YEAR variable for mate-about dialog

rbuj:
Test 1:
```
./autogen.sh --prefix=/usr
cat mate-about/mate-about.h
make && sudo make install
```
Test 2:
```
mkdir builddir
meson setup builddir --prefix=/usr
cat builddir/mate-about/mate-about.h
sudo ninja -C builddir install
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -56,6 +56,9 @@ AC_DEFINE(MATE_MAJOR, [mate_platform], [Define to the major version])
 AC_DEFINE(MATE_MINOR, [mate_minor], [Define to the minor version])
 AC_DEFINE(MATE_MICRO, [mate_micro], [Define to the micro version])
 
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -56,6 +56,9 @@ AC_DEFINE(MATE_MAJOR, [mate_platform], [Define to the major version])
 AC_DEFINE(MATE_MINOR, [mate_minor], [Define to the minor version])
 AC_DEFINE(MATE_MICRO, [mate_micro], [Define to the micro version])
 
+RELEASE_YEAR=2019
```
yetist: Maybe use this will be better?
```
RELEASE_YEAR=`date +%Y`
```

rbuj: @yetist Thanks, done.

PR #419, opened 2020-01-04
mate-desktop-item: Remove escape_string_and_dup/decode_string_and_dup

rbuj:
These functions no longer need to be implemented:
  * escape_string_and_dup -> g_strescape
  * decode_string_and_dup -> g_strcompress


rbuj:
All all non-ASCII chars are also escaped.
```shell
$ ./test-ditem /usr/share/applications/mate-about.desktop
$ cat foo.desktop
```

PR #421, opened 2020-01-12
Backports 2

raveit65:
After `mate_desktop_thumbnail_scale_down_pixbuf` is replaced at eom and caja we don't need to ship it wilth libmate-desktop.
https://github.com/mate-desktop/eom/commit/f7ca9dc7fcc5e088fad181e6f76c5e8a47d70a22
https://github.com/mate-desktop/caja/commit/13d04135dda574bbaa51f56df1783778f423e0f2

Please test and review.


raveit65:
I removed https://gitlab.gnome.org/GNOME/gnome-desktop/commit/3b298f6 from PR, which causes a problem.
Now folder and files icons displayed fast again in caja.
Please test and review.

raveit65:
Thanks for the review.

monsta:
Two API symbols are gone after this, so `LT_VERSION` should be changed I think...

raveit65:
Not sure if this really is needed, you need only a new version of caja and eom which use now `gdk_pixbuf_scale_simple`
https://github.com/mate-desktop/caja/commit/13d04135dda574bbaa51f56df1783778f423e0f2
https://github.com/mate-desktop/eom/commit/f7ca9dc7fcc5e088fad181e6f76c5e8a47d70a22
In other words you can't build <= eom-1.23.1 and <= caja-1.23.2 against latest mate-desktop.
Bumping LT_VERSION doesn't solve this problem, and give us only a lot of hassle with Travis CI configs.

PR #426, opened 2020-02-06
build: remove mate-about.h on distclean target instead of clean

rbuj:


raveit65:
But than `make clean` doesn't remove `mate-about.h`.
So why you want to move it to `distclean`?

rbuj:
before (master): `make clean` removes `mate-about.h`
```shell
$ make distclean; ./autogen.sh --prefix=/usr && make && make clean V=1
<cut>
Making clean in mate-about
make[1]: Entering directory '/home/robert/mate-desktop/mate-about'
 rm -f mate-about
test -z "mate-about.desktop mate-about mate-version.xml mate-about.h" || rm -f mate-about.desktop mate-about mate-version.xml mate-about.h
<cut>
```
after (PR): `make clean` doesn't remove `mate-about.h`
```shell
$ make distclean; ./autogen.sh --prefix=/usr && make && make clean V=1
<cut>
Making clean in mate-about
make[1]: Entering directory '/home/robert/mate-desktop/mate-about'
 rm -f mate-about
test -z "mate-about.desktop mate-about mate-version.xml" || rm -f mate-about.desktop mate-about mate-version.xml
<cut>
```

raveit65:
I know the different between clean and distclean. But again, why you want to move it to `distclean`?
What is the advantage?

rbuj:
The advantage is that the mate-about.h file should be only generated in the configuration step, since it's part of AC_CONFIG_FILES (configure.ac).

rbuj:
$(version_DATA) should be also removed on distclean target instead of clean, i.e. it should be part of DISTCLEANFILES instead of CLEANFILES.

PR #427, opened 2020-02-06
mate-bg: remove warning function declaration isn’t a prototype

rbuj:
```
mate-bg.c:618:1: warning: function declaration isn’t a prototype [-Wstrict-prototypes]
  618 | get_wallpaper_cache_dir ()
      | ^~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/libmate-desktop/mate-bg.c
+++ b/libmate-desktop/mate-bg.c
@@ -615,7 +615,7 @@ mate_bg_get_filename (MateBG *bg)
 }
 
 static inline gchar *
-get_wallpaper_cache_dir ()
+get_wallpaper_cache_dir (void)
 {
 	return g_build_filename (g_get_user_cache_dir(), MATE_BG_CACHE_DIR, NULL);
 }


lukefromdc:
I did some research, and this change is correct so merging:

We have a good explanation of what this change does at 
https://stackoverflow.com/questions/42125/function-declaration-isnt-a-prototype

A prototype is by definition a function declaration that specifies the type(s) of the function's argument(s).

A non-prototype function declaration like

int foo();

is an old-style declaration that does not specify the number or types of arguments. (Prior to the 1989 ANSI C standard, this was the only kind of function declaration available in the language.) You can call such a function with any arbitrary number of arguments, and the compiler isn't required to complain -- but if the call is inconsistent with the definition, your program has undefined behavior.

For a function that takes one or more arguments, you can specify the type of each argument in the declaration:

int bar(int x, double y);

Functions with no arguments are a special case. Logically, empty parentheses would have been a good way to specify that an argument but that syntax was already in use for old-style function declarations, so the ANSI C committee invented a new syntax using the void keyword:

int foo(void); /* foo takes no arguments */

PR #430, opened 2020-02-15
Make the build reproducible

lamby:
Whilst working on the [Reproducible Builds](https://reproducible-builds.org/) effort I noticed that mate-desktop could not be built reproducibly.

This is is because it generated a "release year" based on the current build date which was embedded in the binary and possibly other files. A patch attached that uses [SOURCE_DATE_EPOCH](https://reproducible-builds.org/specs/source-date-epoch/) instead.

As a side-effect, this fixes the (accidental) reset of `MATE_DATE` in the Autotools build system in inherited from `mate-desktop-environment.git`.

This was originally filed in Debian as [#951357](https://bugs.debian.org/951357).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -45,8 +45,10 @@ MATE_PLATFORM=mate_platform
 MATE_MINOR=mate_minor
 MATE_MICRO=mate_micro
 MATE_DATE=`date +"%Y-%m-%d"`
+if test "x$SOURCE_DATE_EPOCH" != "x"; then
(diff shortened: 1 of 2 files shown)


rbuj:
It should also be done on the mate-about/meson.build file, otherwise the build won't be reproducible using meson.

lamby:
Sure thing; updated patch attached.

rbuj:
test: [Year 2038 problem](https://en.wikipedia.org/wiki/Year_2038_problem) PASSED
```
$ SOURCE_DATE_EPOCH=2147617954 ./autogen.sh --prefix=/usr && make && sudo make install
$ mate-about
```

PR #450, opened 2020-08-05
a11y-keyboard: Add capslock-beep-enable gsetting

cwendling:
Add *capslock-beep-enable* gsetting, and bump the version to allow other component to require it.

Required by https://github.com/mate-desktop/mate-settings-daemon/pull/330 and https://github.com/mate-desktop/mate-control-center/pull/583.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,5 @@
+### mate-desktop 1.25.0
+
 ### mate-desktop 1.24.0
 
   * Translations update
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,5 @@
 m4_define([mate_platform], [1])
-m4_define([mate_minor], [24])
+m4_define([mate_minor], [25])
 m4_define([mate_micro], [0])
 
 m4_define(mate_version, [mate_platform.mate_minor.mate_micro]),
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project(
   'mate-desktop',
   'c',
-  version: '1.24.0',
+  version: '1.25.0',
   license: [ 'GPL-2', 'FDL-1.1', 'LGPL-2' ],
 )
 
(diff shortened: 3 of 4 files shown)


NEWS
```diff
@@ -1,3 +1,7 @@
+### mate-desktop 1.25.0
+
+  * a11y-keyboard: Add capslock-beep-enable setting
```
raveit65, line 3: Can you only add new version here?

Adding Commit history will be done with the release commit when releasing.

raveit65:
Version bump is for mate-applets ?

NEWS
cwendling: OK, but without this Travis fails because distcheck fails.

cwendling:
> Version bump is for mate-applets and mcc?

For mcc and msd, yes, as they use the gsetting they require the schema from this PR.

NEWS
raveit65: No i mean add only the version to NEWS.
```
### mate-desktop 1.25.0


### mate-desktop 1.24.0
```
Distcheck requires only a version number.
When this is merged, i will make a new release to make your other PRs happy.

raveit65: For better understanding.
For releasing a tarball i need a minimum change (adding git log entries to NEWS) to generate a commit which can be tagged.
A commit + tag will start the machinery at travis, github and our website to create a new release with tarball.

cwendling: OK got it, I removed the list item.

cwendling:
@raveit65 I plan to submit a second set of changes adding more settings for a related bu separate feature, would you prefer me to submit both so to have it all in one release?

I planned to submit it sequencially because the changes are separate and would conflict if I made two concurrent PRs, but I can as well submit them both in the same PR, or when this mate-desktop one is merged submit the second set before you release.  Just tell me what's best for you :)

raveit65:
Merged :)
I If you have already prepared the second change and it is easy like this one, do another PR before i release 1.25.

cwendling:
Thanks :)  https://github.com/mate-desktop/mate-desktop/pull/451 -- the msd PR will come soon, but I have an unrelated thing to deal with first.

PR #459, opened 2021-01-10
g_settings_get_string always returns a newly-allocated string

rbuj:


raveit65 (approved):
LGTM,
there is a solid black background after session start if caja or m-s-d doesn't handle the desktop.

PR #461, opened 2021-01-20
mate-color-select: Do not use deprecated gtk_button_new_from_stock

rbuj:


lukefromdc (approved):
This works fine, no change in the mate-color-select dialog. This beautiful dialog is now underused in MATE, the current replacement from GTK3 is nowhere near as good and is used in most places right now. Even setting a custom panel background color uses the GTK3 dialog. I  could only find it in caja's "background and emblems" dialog which for some reason on my current caja install only showed "patterns" but worked on an older version of caja. Trying to find the offending commit now

raveit65 (approved):
LGTM,
we need to update transifex resource....

rbuj:
@raveit65 done

PR #466, opened 2021-02-05
Use glib's i18n macros and remove a local macro to init gettext

rbuj:
- use g_print instead of printf which requires a missing include stdio.h


raveit65 (approved):
LGTM,
 builds and runs fine.

PR #470, opened 2021-05-12
MATE About: Fix off-by-one error, display yerba mate comment

thesquash:
The MATE About program is designed to display one of six messages at random
each time you launch the application, showcasing the fundamental values of
the MATE Project.  Unfortunately, due to an off-by-one programming error
introduced many years ago, only the first five messages ever get displayed;
the sixth message, referring to where MATE got its name from, is never chosen.
This was due to a misunderstanding of the GLib function `g_random_int_range`:
The function is and always was designed to be passed an inclusive lower and
an exclusive upper value for the random integer, generating a random integer
that is greater than or equal to the lower value and less than the upper value
specified.  The MATE About application errantly subtracted 1 from the upper
value itself, and thus only messages 1 through 5 were ever picked.

This commit removes the unnecessary subtraction.

### DIFF
--- a/mate-about/mate-about.c
+++ b/mate-about/mate-about.c
@@ -58,7 +58,7 @@ void mate_about_run(void)
      * The comments index must not be more than comments_count - 1
      */
     gtk_about_dialog_set_comments (mate_about_dialog,
-                                   _(comments_array[g_random_int_range (0, comments_count - 1)]));
+                                   _(comments_array[g_random_int_range (0, comments_count)]));
 
     gtk_about_dialog_set_authors(mate_about_dialog, authors);
     gtk_about_dialog_set_artists(mate_about_dialog, artists);


vkareh (approved):
LGTM
Thanks!

lukefromdc:
Just installed this, and I too can confirm it works as expected

thesquash:
Thanks for the testing, but wouldn't you expect this to work?

From https://developer.gnome.org/glib/unstable/glib-Random-Numbers.html#g-random-int-range:

g_random_int_range ()

[gint32](https://developer.gnome.org/glib/unstable/glib-Basic-Types.html#gint32)
g_random_int_range (
[gint32](https://developer.gnome.org/glib/unstable/glib-Basic-Types.html#gint32)
begin

,
[gint32](https://developer.gnome.org/glib/unstable/glib-Basic-Types.html#gint32)
end

);

Returns a random [gint32](https://developer.gnome.org/glib/unstable/glib-Basic-Types.html#gint32) equally distributed over the range [begin ..end -1].

Parameters

begin

lower closed bound of the interval

end

upper open bound of the interval

lukefromdc:
Expecting is one thing, testing and making sure is another. Expect 20 things to work,make one mistake and that could be the next big exploiit. 

While the rotating list of messages in mate-about does not take user input and thus would be difficult to exploit, there are lots of places im computing in general where a simple bug (whicj this off by one is) can be a real problem. 

They just closed a nasty one in sudo where running sudo -s with a special argument list ending in a backslash gives root access, presumably without the sudo passphrase. I don't know what did this, but an off by one in computing the length of the string submitted by a user could theoretically do something like this by causing a buffer overflow.

Imagine someone at a motor vehicle administration drivers license renewal kiosk using Linux after one too many Windows machinea got hacked. Now imagine that sudo bug being invoked from say, the panel command applet. User is now root on the whole motor vehicle administration computer system. That means read/write access to other people's driving records. Now we get fake traffic tickets or even fake warrants against whoever the attacker doesn't like.Lots of "fun" by the side of the road with nasty armed cops.

Starting with what you know works save both you and all testers time, but is not the same as being sure. Even static analysis tools can be used out of context or otherwise mislead a programmer.

Remember that SSL key blacklist issue? That began whan static analysis detected use of uninitialized ram, so programmers commented out the "offending" line. Trouble is, that uninitialized ram was where all the randomneas came from in key generation. The resulting keys could be easily brute-forced. Globally thay can mean dissidents being rounded up after their communications were read. To this day all those keys have to be blacklisted.

PR #472, opened 2021-07-17
Fix compilation warning

zhuyaliang:
I fixed some compilation warnings  replaced some outdated API.
Next I'll use ```gdk_raba``` replaces ```gdk_color```

### DIFF
--- a/libmate-desktop/mate-colorbutton.c
+++ b/libmate-desktop/mate-colorbutton.c
@@ -260,7 +260,10 @@ draw (GtkWidget      *widget,
 {
   MateColorButton *color_button = MATE_COLOR_BUTTON (data);
   cairo_pattern_t *checkered;
+  GtkStyleContext *context;
+  GdkRGBA  rgba;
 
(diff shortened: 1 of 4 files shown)


lukefromdc (commented):
We get a new runtime warning with this:

`(mate-color-select:144364): Gtk-CRITICAL **: 13:47:22.095: gtk_widget_grab_default: assertion 'gtk_widget_get_can_default (widget)' failed`

Note that (as in master) the mate-color-select comes up with white (FFFFFF) as the default color, so don't be confused by the color value and right half of the color strip not changing as the triangulalr hue widget is rotated until the saturation circle is moved off of pure white. The runtime warning has nothing to do with that, as confirmed by testing master as well

zhuyaliang:
@lukefromdc Repair the warning generated at run time, please test again

lukefromdc (approved):
The runtime warning is gone, the built mate-color-select binary works fine

vkareh (approved):
LGTM

PR #473, opened 2021-07-18
Replace the obsolete API in mate-colorsel.c file

zhuyaliang:
Replace the obsolete API in mate-colorsel.c file 
Use ```gdk_seat``` and ```gdk_device``` replace ```gdk_display```
I can't provide a test program to test whether this change is correct

### DIFF
--- a/libmate-desktop/mate-colorsel.c
+++ b/libmate-desktop/mate-colorsel.c
@@ -1628,6 +1628,17 @@ make_picker_cursor (GdkScreen *screen)
   return cursor;
 }
 
+static GdkDevice *get_device (GdkDisplay *display)
+{
+  GdkSeat   *seat;
+  GdkDevice *device;
+
+  seat = gdk_display_get_default_seat (display);
+  device = gdk_seat_get_pointer (seat);
+
+  return device;
+}
+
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
The color select widget in Caja's "backgrounds and emblems" dialog uses this, and runs without any issues or runtime warnings on a second instance of caja as root (to avoid just connecting to existing desktop instance of caja as normal user).

I see we are basically replacing old code for getting the pointer and keyboard here. Both work fine in the test above

lukefromdc:
Note that when mate-panel (and probably other instances in MATE) was ported to GTK3, what was then GTK2's color select dialog (and became the MATE color select dialog) was being used for the panel background custom color selector, so that had to be ported to GTK3's color selection dialog. We need to port it back to  caja_color_selection_dialog which is now part of mate-desktop rather than GTK, and do the same for caja/mate-settings-daemon desktop background color selector, the color selectors in mate-system-monitor, etc.

NOT a high priority job, but something to get on when there is time. I played with it a bit tonight but found it was more complex than I expected

vkareh (approved):
LGTM

PR #489, opened 2021-10-19
Fix versions

rezso:
mate_minor_version and mate_micro_version always 1. In addition, mate_minor / mate_minor_version and mate_micro / mate_micro_version always same.


raveit65 (approved):
LGTM,
thanks.

PR #490, opened 2021-10-22
mate-desktop-thumbnail: fix memory leak

rbuj:


lukefromdc (approved):
Wasn't sure what this thumbnails so checked all possibities: desktop background thumbnails, image thumbnails, video thumbnails,  and the mate-appearance-properties theme thumbnails. All worked same as before with this installed. Also, if this did NOT work, I would expect a segfault as soon as something tried to use the freed variable, so this cannot be setting up a use after free.

PR #495, opened 2021-11-27
mate-bg: Dereference of null pointer

rbuj:


lukefromdc (approved):
Builds fine, and a restarted caja changes backgrounds without issue once this is installed.Simple check for NULL, no new issues should come from this simple fix.

PR #525, opened 2022-07-25
configure.ac: The macro 'AC_HELP_STRING' is obsolete

rbuj:


lukefromdc:
Once again, the new code this depends on has been around over a decade,I don't think MATE 1.28 will ever ship with an LTS that first shipped before 2010

lukefromdc (approved):
While I normally build mate-desktop with meson/ninja, a test configure and build with this worked fine, and no problems using ./configure --help after initial ./autogen.sh either.

Will merge, AS_HELP_STRING should be available on any LTS that would ever update to forthcoming MATE 1.28

PR #529, opened 2022-07-28
Do not intersect time_t with gint64

rbuj:


lukefromdc:
What is meant here by "intersect?"

lukefromdc (commented):
Builds and runs fine,though I do not know how to properly test this

lukefromdc (approved):
Did some digging,looks like we are getting rid of an old variable type not cleanly defined in C:

https://en.cppreference.com/w/c/chrono/time_t

"time_t
 
C
 
Date and time utilities
 
Defined in header <time.h>
		
typedef /* unspecified */ time_t;
		
		

Arithmetic (until C11) Real (since C11) type capable of representing times.

Although not defined by the C standard, this is almost always an integral value holding the number of seconds (not counting leap seconds) since 00:00, Jan 1 1970 UTC, corresponding to POSIX time. "

Issue #246, opened 2017-01-19
Please add an option to set the max volume above 100%


michaelmrose:
Some combinations of hardware + source are too quiet at 100% Adjusting the sound above 100% requires right clicking on the sound applet selecting properties adjusting it to the desired spot then if you use the normal volume control keys at this point will put the sound below the 100% threshold once again requiring the above procedure.  Rather than having a fixed boundry at  100% it would be nice if there was an option for this.

raveit65:
Above 100% for a source means the sound is distorted for a normal input.

michaelmrose:
For some inputs read movies with highly variable sound levels with the sound turned up to 100% the quiet parts are too low to understand conversation and the only practical way to enjoy some movies at all on the machine is to turn the quiet parts up higher.  This doesn't sound remotely distorted in this case.

My fix was to bind the volume up/down keys to scripts that turn the volume up/down in 5% increments and have it actually max out at 130 instead of 100%.  This misses out on the visual feedback of the volume meter that pops up and its probably unusable to 90% of people using mint. Lacking an option to set the max it would perhaps still be nicer if it went up a bit past 100%.  If it sounds bad to people they just wont turn it up that much.

michaelmrose:
Some other comments regarding turning audio past 100%

From http://askubuntu.com/questions/41348/why-can-the-volume-go-higher-than-100

"Overamplifying a volume is a good feature, especially for laptops where internal speakers are not loud. I was using this feature since Ubuntu (9.10) on my FujitsuSiemens SI1520. It helped me very much to listen to music."

"I use it when i listen to speeches that were recorded at rather low volume. It might distort music/sound effects etc but it is rather useful in those cases when you just can't listen because of bad recording."

From http://superuser.com/questions/146784/sound-volume-increase-beyond-100-whenever-possible-on-linux

"Some audio output from files or streams is too low. It is obvious that hardware is able to play the same sounds but louder but because of the data it just plays it at some low level even at 100% volume."

Should I dig up about a hundred more?

flexiondotorg:
You can exceed 100% volume in the Sound Properties.

michaelmrose:
[...]

PR #530, opened 2022-07-29
Add setting for adjustment of audio volume above 100 per cent:  Part 1

thesquash:
There is often a need for the user to increase the audio playback volume above
the volume level known as "100% volume".  While increasing the audio volume
above 100% can result in degraded audio quality, sometimes the audio was, for
example, originally recorded at an extremely low volume, and the user has no
other option to clearly hear the audio.  Unfortunately, most MATE applications
with volume controls do not allow the user to set the volume level above 100%.
For example, the main MATE Sound Preferences dialog lets you set the audio
volume beyond 100% (when possible), whereas the Volume Control Applet, Volume
Control status icon, and special "multimedia" volume control keys do not.  In
fact, if the user even tries to change the volume using any of the latter
methods, and the current volume level is above 100%, these latter methods will
all reduce the volume to 100%, even if the user tried to increase the volume!

This is part 1 of a patch to change this situation.  All that this patch does
is add a new, standard MATE setting which determines whether the user should
be allowed to set the audio volume above 100%.  The rest of the functionality
of this patch is in other commits / pull requests.

Part 2:  https://github.com/mate-desktop/mate-media/pull/191
Part 3:  https://github.com/mate-desktop/mate-settings-daemon/pull/389

When all three parts are applied, https://github.com/mate-desktop/mate-desktop/issues/246 can be closed.

### DIFF
--- a/schemas/org.mate.sound.gschema.xml
+++ b/schemas/org.mate.sound.gschema.xml
@@ -30,5 +30,17 @@
       <summary>Input feedback sounds</summary>
       <description>Whether to play sounds on input events.</description>
     </key>
+    <key name="volume-overamplifiable" type="b">
+      <default>false</default>
+      <summary>Permit the output volume to exceed 100%</summary>
+      <description>
+        If this is set to TRUE, the user will be allowed to set the output /
+        speaker volume to above 100%.  By setting the output volume above
+        100%, the sound server will amplify the audio above what the audio
+        hardware can do cleanly.  This can lead to audio quality degradation,
+        but may be useful for listening to extremely quiet audio recordings,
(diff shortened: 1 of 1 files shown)


rbuj (approved):
LGTM. It's intended to be bound to the property "extended" of GvcChannelBar.
```C
        properties[PROP_EXTENDED] =
                g_param_spec_boolean ("extended",
                                      "Extended",
                                      "Allow the scale to be extended above normal volume",
                                      FALSE,
                                      G_PARAM_READWRITE |
                                      G_PARAM_STATIC_STRINGS);
```

raveit65 (approved):
LGTM,
toogling gesettings key works fine.

PR #538, opened 2022-11-07
Revert "[mate-bg] small cleanup"

cwendling:
It is not possible to use the `list` pointer after it has been deleted, so the "cleanup" this commit made lead to using freed memory if any item actually got clean up.

This "cleanup" also don't seem meaningful to me, as all it does otherwise is trade an assignation for a redundant test -- either of which the compiler might happily optimize out.

This reverts commit 47426c90d10e9f738ecf89f35db94ca8deff55e0.

### DIFF
--- a/libmate-desktop/mate-bg.c
+++ b/libmate-desktop/mate-bg.c
@@ -2002,19 +2002,18 @@ static gboolean
 blow_expensive_caches (gpointer data)
 {
 	MateBG *bg = data;
-	GList *list;
+	GList *list, *next;
 
 	bg->blow_caches_id = 0;
 
-	if (bg->file_cache) {
-		for (list = bg->file_cache; list != NULL; list = list->next) {
-			FileCacheEntry *ent = list->data;
+	for (list = bg->file_cache; list != NULL; list = next) {
+		FileCacheEntry *ent = list->data;
+		next = list->next;
 
-			if (ent->type == PIXBUF) {
-				file_cache_entry_delete (ent);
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
I don't get any problems with caja or with changing backgrounds with or without this applied. Maybe a compiler difference means the change being reverted works for some but not all?  Anyway, a use-after-free (even if some compilers optimize it out) is at least a crasher and at worst a potential security risk if something sensitive got written to the freed RAM by another process between freeing it and using it after free.

lukefromdc:
By any chance do we leak` next` (declared as glist) with this change?

cwendling:
@lukefromdc I don't think it's possible a compiler would "fix" this code, but I think it's likely there will be nothing to cleanup on most systems.  I'm not entirely sure what are the rules for the cleanup here, but in my case I re-enabled desktop icons after having them off for a long long time, I'm not even sure when was the last time I did run Caja itself (t's my dev machine and I manage files there mostly on the terminal, and didn't have icons on the desktop, so I open Caja only occasionally there), so maybe it's just that in my case some things caches or so got so old the cleanup kicked in?
Anyway, I'm totally sure the code had incorrect logic, which lead to a crash on my end.

> By any chance do we leak `next` (declared as glist) with this change?

No, it's just a stack-allocated pointer to the next element because we might free the current one inside the loop, hence not permitting dereferencing `list` at the end of the loop (well, in the "post loop" section of the `for` construct) -- which was the crasher bug I encountered.

Thanks for the review & merge :)

Issue #550, opened 2023-03-29
Add a new ImageMenu API


zhuyaliang:
```gtk_image_menu_item_new``` has been deprecated since version 3.10 and should not be used in newly-written code.We use a lot of ```gtk_image_menu_item``` in our code, which has resulted in a large number of compilation warnings. To eliminate compilation warnings, I plan to add a new API ```mate_image_menu_item```.
If there is no objection, I will start working immediately

lukefromdc:
Go for it, being aware we probably won't ever go to GTK 4 for many reasons at this point. Taking
this in-house would mean we are no longer vulnerable to bitrot on this item though. Right now we
have GNOME responsible for maintaining widgets and functions they no longer use in-house,
which means those who write and maintain them have to deliberately test on code they use for
nothing else if they wish to prevent them from bitrotting.

PR #551, opened 2023-03-30
Add mate image menu item apis

zhuyaliang:
Add ```mate_image_menu_item``` apis to replace ```gtk_image_menu_item``` to remove deprecated compilation warnings
After this PR is merged, I will gradually replace the deprecated ```gtk_image_menu_item``` in other projects

#550 

## test
run ```test-image-menu-item``` command

### DIFF
--- a/docs/reference/mate-desktop/mate-desktop-docs.sgml
+++ b/docs/reference/mate-desktop/mate-desktop-docs.sgml
@@ -41,6 +41,11 @@ The libmate-desktop library contains APIs that can be useful for a few applicati
     <xi:include href="xml/mate-languages.xml"/>
   </part>
 
+  <part id="image-menu-item">
+    <title>MateImageMenuItem</title>
+    <xi:include href="xml/mate-image-menu-item.xml"/>
+  </part>
+
   <part id="libmate">
     <title>Miscellaneous</title>
     <xi:include href="xml/mate-desktop-item.xml"/>
--- a/docs/reference/mate-desktop/mate-desktop-sections.txt
+++ b/docs/reference/mate-desktop/mate-desktop-sections.txt
@@ -280,6 +280,16 @@ mate_language_has_translations
 </SECTION>
 
 <SECTION>
+<FILE>mate-image-menu-item</FILE>
+<TITLE>MateImageMenuItem</TITLE>
+mate_image_menu_item_new
+mate_image_menu_item_new_with_label
+mate_image_menu_item_new_with_mnemonic
+mate_image_menu_item_set_image
+mate_image_menu_item_get_image
+</SECTION>
+
+<SECTION>
 <INCLUDE>mate-desktop-utils.h</INCLUDE>
 <FILE>mate-desktop-utils</FILE>
 <TITLE>Miscellaneous Functions</TITLE>
--- a/libmate-desktop/Makefile.am
+++ b/libmate-desktop/Makefile.am
@@ -15,7 +15,8 @@ libmate_desktop_HEADERS = \
 	mate-colorbutton.h \
 	mate-colorsel.h \
 	mate-hsv.h \
-	mate-colorseldialog.h
+	mate-colorseldialog.h \
+	mate-image-menu-item.h
 
 lib_LTLIBRARIES = libmate-desktop-2.la
 
@@ -31,7 +32,7 @@ AM_CPPFLAGS =							\
 
 AM_CFLAGS = $(WARN_CFLAGS)
 
-noinst_PROGRAMS = test-desktop-thumbnail test-ditem test test-languages
+noinst_PROGRAMS = test-desktop-thumbnail test-ditem test test-languages test-image-menu-item
 
 CLEANFILES =
 
@@ -52,7 +53,8 @@ introspection_sources = \
 	mate-colorsel.c \
 	mate-hsv.c \
 	mate-colorseldialog.c \
-	edid-parse.c
+	edid-parse.c \
+	mate-image-menu-item.c
 
 libmate_desktop_2_la_SOURCES =		\
 	$(introspection_sources)	\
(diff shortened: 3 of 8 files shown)


raveit65:
Test binary seems to be working.
test-image-menu-item
Menu looks like before an icons are there.

raveit65:
I guess `mate-image-menu-item.c` is based on deprecated `gtk_image_menu_item` ?

raveit65 (approved):
I didn't compare mate-image-menu-item.c with deprecated gtk_image_menu_item but code seems to be OK, and i didn't test meson build.

raveit65:
I didn't noticed that first, but why the space between icon and text from `edit-copy` is not the same like other menu entries in my screenshot?

lukefromdc (approved):
Just got successful builds with both meson and autotools. In both cases the test binary worked fine. Looks like a nice reimplementation of gtk_image_menu_item and doing this means these functions get maintained by folks that actually use them. It's a lot to ask GNOME fpr them to permanently be responsible for testing code they stopped using in their own projects years ago.

lukefromdc:
Will wait on merge to see if the icon space issue gets a fix so that doesn't have to be a whole separate PR.

zhuyaliang:
@raveit65 @lukefromdc 
```edit copy``` icon using ```gtk_container_add (GTK_CONTAINER (menuitem), label)``` This will introduce new spacing, as is the case with GTK, which is normal.
```
    menuitem = mate_image_menu_item_new (); 
    label = gtk_label_new ("test image_menu_item_new");
    gtk_container_add (GTK_CONTAINER (menuitem), label);
```

**Comment by lukefromdc** [comment]
This looks like 
`mate_image_menu_item_new ();`
is spaciing the text farther from the icon than 
`mate_image_menu_item_new_with_label`
and
`mate_image_menu_item_new_with_mnemonic
`
Which would affect all uses of this code and needs to be fixed in mate-image-menu-item.c

**Comment by lukefromdc** [comment]
Ahh-I see we are not using the same code all three test examples

**Comment by lukefromdc** [comment]
Looks like use of `mate_image_menu_item_new` forces use of ` gtk_container_add ` and the other two do not. If this is the source of the extra space the question becomes this: would actual uses mix these or not, and is that much space acceptable given it's about double what usually comes up with the older deprecated code

**Comment by zhuyaliang** [comment]
@lukefromdc The code is completely from ```GTK3```, and I extracted the code we need from it.
GTK3 provides the following apis：
```
void            gtk_image_menu_item_set_image ()
GtkWidget *	gtk_image_menu_item_get_image ()
GtkWidget *	gtk_image_menu_item_new ()
GtkWidget *	gtk_image_menu_item_new_from_stock ()
GtkWidget *	gtk_image_menu_item_new_with_label ()
GtkWidget *	gtk_image_menu_item_new_with_mnemonic ()
gboolean        gtk_image_menu_item_get_use_stock ()
void            gtk_image_menu_item_set_use_stock ()
gboolean        gtk_image_menu_item_get_always_show_image ()
void            gtk_image_menu_item_set_always_show_image ()
void            gtk_image_menu_item_set_accel_group ()
```
We only kept what we needed
```
GtkWidget*      mate_image_menu_item_new ()
GtkWidget*      mate_image_menu_item_new_with_label ()
GtkWidget*      mate_image_menu_item_new_with_mnemonic ()
void            mate_image_menu_item_set_image ()
GtkWidget*      mate_image_menu_item_get_image ()
```
Three types of ```image_menu_item``` are tested in the test code.The same behavior exists with the older deprecated GTK code，The test results are the same, we should merge as is

**Comment by raveit65** [comment]
I never noticed in real gtk menus such space issue, so it seem that the 3 cases are never used in the same menu.
We have to see how it looks in real menus. Code builds fine, so let's merge it.
After merge i can do  a new release which needs to be add to travis-ci config of depending packages.

**Comment by raveit65** [comment]
I just release 1.27.1.
Here is an example what needs to be add to `.build.yml` in `before_scripts` part of a depending package.
https://github.com/mate-desktop/mate-control-center/blob/master/.build.yml#L155
```
  - MATE_DESKTOP_VERSION=1.25.0

before_scripts:
  - cd ${START_DIR}
  - if [ ! -f mate-desktop-${MATE_DESKTOP_VERSION}.tar.xz ];then
  -     curl -Ls -o mate-desktop-${MATE_DESKTOP_VERSION}.tar.xz https://github.com/mate-desktop/mate-desktop/releases/download/v${MATE_DESKTOP_VERSION}/mate-desktop-${MATE_DESKTOP_VERSION}.tar.xz
  - fi
  - tar xf mate-desktop-${MATE_DESKTOP_VERSION}.tar.xz
  - cd mate-desktop-${MATE_DESKTOP_VERSION}
  - if [ ${DISTRO_NAME} == "debian" -o ${DISTRO_NAME} == "ubuntu" ];then
  -     ./autogen.sh --prefix=/usr --libdir=/usr/lib/x86_64-linux-gnu --libexecdir=/usr/lib/x86_64-linux-gnu
  - else
  -     ./autogen.sh --prefix=/usr
  - fi
  - if [ ${TRAVIS} == "false" ]; then
  -     make clean
  - fi
  - make
  - make install
 ```

raveit65:
@zhuyaliang 
Can you add `mate_image_menu_item_get_always_show_image ()` and `gtk_image_menu_item_get_use_stock ()`, please?
Atril use it:
```
shell/ev-window.c:7945:	gtk_image_menu_item_set_always_show_image (GTK_IMAGE_MENU_ITEM (menuitem), TRUE);
shell/ev-window.c:7948:	gtk_image_menu_item_set_always_show_image (GTK_IMAGE_MENU_ITEM (menuitem), TRUE);
shell/ev-sidebar-links.c:337:	item = gtk_image_menu_item_new_from_stock ("gtk-print", NULL);
```

zhuyaliang:
@raveit65 
The 'always_show_image' Properties can be set in the UI file,```gtk_image_menu_item_new_from_stock``` can be used ```mate_image_menu_item_new_label``` and ```mate_image_menu_item_set_image``` replace

zhuyaliang:
@raveit65 
https://github.com/mate-desktop/mate-desktop/pull/551/files#diff-cec4eb46031a7bc2622688df8cdd920f921e46fe18f75995cbea8e683ae7f503R527-R530
I made some changes while porting the code, and the ```visible``` here has been fixed to ```TRUE```,That is to say, if ```mate_image_menu_item_set_image```  is used, there is no need to use ```gtk_image_menu_item_set_always_show_image``` again
```
g_object_set (image,
                    "visible", TRUE,
                    "no-show-all", TRUE,
                     NULL);
```

Issue #574, opened 2023-09-07
Add a mate-portals.conf for xdg-desktop-portal


smcv:
As a small and straightforward step towards mate-desktop/mate-desktop#504, it would be useful for some central component of MATE to install a `mate-portals.conf` for xdg-desktop-portal >= 1.17, indicating the MATE maintainers' recommended portal backend for use with MATE. If that recommendation is currently [x-d-p-gtk](https://github.com/flatpak/xdg-desktop-portal-gtk), then that would be as simple as this (untested, I don't normally run MATE):

```
# /usr/share/xdg-desktop-portal/mate-portals.conf
[preferred]
default=gtk;
```

[More information](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1050803)

If this file doesn't exist, x-d-p will fall back to the deprecated `UseIn` keys.

The GNOME equivalent of this is [in gnome-session](https://gitlab.gnome.org/GNOME/gnome-session/-/merge_requests/95). In MATE, I've guessed that mate-session-manager would similarly be the most appropriate home for this, as the component that sets `DesktopNames=MATE`; but any component that mate-session-manager depends on (like mate-desktop?) would also be fine.

Each x-d-p backend can implement any subset of the interfaces used by x-d-p: they are not necessarily required to implement the full set. It's entirely possible that the right thing to do for MATE would be to rely on xdp-gtk for the interfaces that are straightforward to implement in a mostly-desktop-independent way (like the file chooser), but then use a MATE-specific backend as requested in mate-desktop/mate-desktop#504 for other interfaces (like maybe screenshots). For example, GNOME uses xdp-gnome for most interfaces, but the backend for Secrets is in gnome-keyring, and the backend for Access is in gnome-shell, both of which were easier or more appropriate places to put that functionality.

In mate-desktop/mate-desktop#504, @raveit65 writes:
> Nope, xapps are cinamon only production. There xapps are only forks from origin MATE applications like Atril, eom, engrampa. In other words we don't need that.

x-d-p-xapp seems to think that it's targeting Cinnamon, MATE and XFCE, not just Cinnamon - it has `UseIn=X-Cinnamon;MATE;XFCE;`. This is one of the reasons why x-d-p added `portals.conf(5)` - so that it can be the desktop environment, not the portal implementation, that makes the choice of which portal implementation(s) each desktop should be using.

#### Steps to reproduce the behaviour

[...]

PR #577, opened 2023-09-14
add xdg-desktop-portal config file

raveit65:
fixes https://github.com/mate-desktop/mate-desktop/issues/574

This installs a mate-portals.conf for xdg-desktop-portal >= 1.17, indicating that MATE recommended portal backend for use with MATE. Recommendation is currently [x-d-p-gtk](https://github.com/flatpak/xdg-desktop-portal-gtk).
Edit:
The mate.portals.conf file is a replacement for the UseIn key inside the portal implementation files; it is also a replacement for people disabling unwanted portal support.

### DIFF
--- a/xdg-portal/Makefile.am
+++ b/xdg-portal/Makefile.am
@@ -1,0 +1,5 @@
+NULL =
+
+xdgportaldir = $(datadir)/xdg-desktop-portal
+dist_xdgportal_DATA = mate-portals.conf
+
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,4 +1,4 @@
-SUBDIRS = libmate-desktop man docs schemas tools icons po
+SUBDIRS = libmate-desktop man docs schemas tools icons po xdg-portal
 
 if MATE_ABOUT_ENABLED
 SUBDIRS += mate-about
--- a/configure.ac
+++ b/configure.ac
@@ -257,6 +257,7 @@ schemas/org.mate.background.gschema.xml
 man/Makefile
 tools/Makefile
 icons/Makefile
+xdg-portal/Makefile
 ])
 
 AC_OUTPUT
--- a/meson.build
+++ b/meson.build
@@ -104,6 +104,7 @@ endif
 subdir('po')
 subdir('tools')
 subdir('schemas')
+subdir('xdg-portal')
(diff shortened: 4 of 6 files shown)


raveit65:
Adding `org.freedesktop.impl.portal.Secret=gnome-keyring;` is fine because we are using gnome-keyring.

xdg-portal/Makefile.am
```diff
@@ -0,0 +1,8 @@
+NULL =
+
+xdgportaldir = $(datadir)/xdg-desktop-portal
+xdgportal_DATA = mate-portals.conf
```
smcv, line 4: If you changed this to

```suggestion
dist_xdgportal_DATA = mate-portals.conf
```

then you wouldn't need to put it in `EXTRA_DIST`, if I remember correctly how Autotools works.

smcv (approved):
This looks like a good initial implementation, and does what I asked for in #574.

xdg-portal/Makefile.am
raveit65: Yeap that works, `make distcheck` creates a proper tarball with `xdgportaldir`

lukefromdc:
Note that that will work only on autotools builds

raveit65:
> Note that that will work only on autotools builds

I checked meson build with `meson dist --allow-dirty -C build` and did a meson test installation to /usr/local.
/usr/share/xdg-desktop-portal/mate-portals.conf will be installed with meson at my box.

So , what is wrong?

lukefromdc (commented):
I didn;t get any suprises with this, but also didn't get any change in GIMP 3.0 (GTK3, should be able to run in native wayland but didn't check) wayland color picker behavior. That said, I don't use snaps or flatpacks at all, so I don't know much about portals

lukefromdc:
CORRECTION: meson.build IS included so should work either way, Cannot test for function though as I have nothing using portals

raveit65:
We need xdg-desktop-portal >= 1.17 to get this work.
I couldn't reproduce the gimp issue for my self fir fedora 38.
Last posts in gimp report shows that it was an issue with 
xdg-desktop-portal < 1.16
xdg-desktop-portal-gtk < 1.14.1
So i would never see the issue in fedora 38.

PR #581, opened 2023-10-05
mate-dmi: add new api to get dmi info

yetist:


lukefromdc:
What is dmi info? This is the first I've seen that term

zhuyaliang (commented):
Systemd provides the same interface functionality

zhuyaliang:
2023-10-06 08-41-32 创建的截图
We can directly use the ```gdbus``` interface provided by ```systemd```, otherwise it will increase our maintenance costs

lukefromdc:
How would I test this?

yetist:
> We can directly use the ```gdbus``` interface provided by ```systemd```, otherwise it will increase our maintenance costs

Good, thanks.

yetist:
> How would I test this?

you can run the test-dmi command to see the output, but I think we dont need this PR now.

PR #582, opened 2023-10-07
Remove useless mate_color_button apis

zhuyaliang:
The ```mate_color_button``` has not been called by any code. ```GtkColorChooser``` has been used instead of ```MateColorButton```  In order to reduce maintenance costs, we should streamline the code and remove some unused code


raveit65:
It think last usage was with mate-color-chooser but it was converted to gtk-color-chooser some years ago.
https://github.com/mate-desktop/mate-panel/pull/464

zhuyaliang:
The ```mate_color button``` is no longer in use

raveit65:
`make distcheck` runs fine.
For some reason `meson dist` works only with `--allow-dirty` flag.
No idea, it that is a problem.
I noticed that with another repo too.

raveit65 (approved):
LGTM

zhuyaliang:
@raveit65 Can I merge the code? My next PR to replace expired functions and fix compilation warnings will be based on this
@lukefromdc @cwendling @yetist Please help review the code

yetist (approved):
LGTM

lukefromdc (approved):
Both autotools and meson builds finished fine, no surprises on installing the autotools build but none expected as this just removes the unused colorbutton. EDIT: this removes the button API but we keep the stanalone mate-color-select binary

raveit65:
@zhuyaliang @lukefromdc 
During testing new PRs i noticed that `mate-desktop.abi` isn't in EXTRA_DIST or released Tarballs.
Do we don't need this file for runtime?
Is it only for documentation?

raveit65:
Maybe it is useful for users when we ship those file with tarballs?
I can do that.

PR #584, opened 2023-10-08
mate-colorsel: Fix compilation warnings use rgba replace color

zhuyaliang:
Replace ```GdkColor``` with ```GdkRGBA```


lukefromdc (approved):
I get only deprecation warnings rebuilding with this applied after removing the button API's. Note: ONLY these 4 files dropped in so warnings from other files would not be shown.

 Nice that we do in fact both keep and maintain the standalone mate-color-select binary

raveit65 (approved):
LGTM,
- no new build warnings
- mate-color-select runs fine

raveit65:
Anyone else like to review this PR?
Because it is a huge code change or should we merge it?

lukefromdc:
I'm OK with merging it but will leave the call to you on how long to wait for any more reviews

libmate-desktop/mate-colorsel.c
```diff
@@ -2239,25 +2222,28 @@ palette_change_notify_instance (GObject    *object,
...
+  GValue value = G_VALUE_INIT;
   gchar *str;
 
+  g_value_init (&value, G_TYPE_STRING);
+
   str = mate_color_selection_palette_to_string (colors, n_colors);
+  g_value_set_string (&value, str);
 
-  gtk_settings_set_string_property (gtk_settings_get_for_screen (screen),
+  g_object_set_property (G_OBJECT (gtk_settings_get_for_screen (screen)),
```
cwendling, line 2244: Any reason not to use `g_object_set()` instead?  I didn't check but I believe you leak the string set into `value`, and although that's easily fixable `g_object_set()` is a lot easier to use in C:
```c
g_object_set (gtk_settings_get_for_screen (screen), "gtk-color-palette", str, NULL);
```

No need for messing with `GValue`s manually.

libmate-desktop/mate-colorsel.c
```diff
@@ -2826,9 +2818,9 @@ mate_color_selection_palette_to_string (const GdkColor *colors,
 
       strs[i] =
         g_strdup_printf ("#%2X%2X%2X",
-                         colors[i].red / 256,
-                         colors[i].green / 256,
-                         colors[i].blue / 256);
+                         (unsigned int)colors[i].red / 256,
+                         (unsigned int)colors[i].green / 256,
+                         (unsigned int)colors[i].blue / 256);
```
cwendling, line 2823: Is this correct?  I would have expected to have to do `colors[i].red * 255` given `GdkRGBA` uses `0..1` ranges (and ofc same for the other channels).

libmate-desktop/mate-colorsel.c
zhuyaliang: done

libmate-desktop/mate-colorsel.c
zhuyaliang: done

libmate-desktop/mate-colorsel.c
```diff
@@ -1022,13 +1025,16 @@ palette_paint (GtkWidget    *drawing_area,
...
     return;
 
   gtk_widget_get_allocation (drawing_area, &allocation);
 
-  gdk_cairo_set_source_color (cr, &gtk_widget_get_style (drawing_area)->bg[GTK_STATE_NORMAL]);
+  palette_get_color (drawing_area, color);
+  GdkRGBA bg = {color[0], color[1], color[2], color[3]};
+  gdk_cairo_set_source_rgba (cr, &bg);
   cairo_rectangle (cr, 0, 0, allocation.width, allocation.height);
   cairo_fill (cr);
```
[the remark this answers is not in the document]
zhuyaliang: done

libmate-desktop/mate-colorsel.c
```diff
@@ -1022,13 +1025,15 @@ palette_paint (GtkWidget    *drawing_area,
...
   GtkAllocation allocation;
+  GdkRGBA bg;
 
   if (gtk_widget_get_window (drawing_area) == NULL)
     return;
 
   gtk_widget_get_allocation (drawing_area, &allocation);
 
-  gdk_cairo_set_source_color (cr, &gtk_widget_get_style (drawing_area)->bg[GTK_STATE_NORMAL]);
+  gtk_style_context_get_color (gtk_widget_get_style_context(drawing_area),GTK_STATE_FLAG_NORMAL, &bg);
```
cwendling, line 1035: you're querying the foreground color, whereas the previous code used the background.

libmate-desktop/mate-colorsel.c
```diff
@@ -1226,7 +1230,7 @@ override_background_color (GtkWidget *widget,
 
   provider = gtk_css_provider_new ();
 
-  css = g_strdup_printf ("* { background-color: %s;}",
+  css = g_strdup_printf ("* { color: %s;}",
```
cwendling, line 1233: ah, maybe coupled with that it works…  but why override with CSS then manually draw it?  it'd be simpler just to paint a given color, wouldn't it?  And if it's more complex, this will change behavior.

libmate-desktop/mate-colorsel.c
```diff
@@ -1929,17 +1902,22 @@ hex_changed (GtkWidget *hex_entry,
     return;
 
   text = gtk_editable_get_chars (GTK_EDITABLE (priv->hex_entry), 0, -1);
-  if (gdk_color_parse (text, &color))
+  if (!g_str_has_prefix (text, "#"))
     {
-      priv->color[COLORSEL_RED] = CLAMP (color.red/65535.0, 0.0, 1.0);
-      priv->color[COLORSEL_GREEN] = CLAMP (color.green/65535.0, 0.0, 1.0);
-      priv->color[COLORSEL_BLUE] = CLAMP (color.blue/65535.0, 0.0, 1.0);
+      text = g_strconcat ("#", text, NULL);
```
cwendling, line 1907: memory leak:
```suggestion
      gchar *tmp = g_strconcat ("#", text, NULL);
      g_free (text);
      text = tmp;
```

libmate-desktop/mate-colorsel.c
```diff
@@ -2756,8 +2740,8 @@ mate_color_selection_palette_from_string (const gchar *str,
               done = FALSE;
             }
 
-          retval = g_renew (GdkColor, retval, count + 1);
-          if (!gdk_color_parse (start, retval + count))
+          retval = g_renew (GdkRGBA, retval, count + 1);
+          if (!gdk_rgba_parse (retval + count, start))
```
cwendling, line 2744: Do you need to handle the case when there's no `#` prefix like done above?

libmate-desktop/mate-colorsel.c
cwendling: What about making things simpler like this?
```diff
diff --git a/libmate-desktop/mate-colorsel.c b/libmate-desktop/mate-colorsel.c
index 148e429..7213c91 100644
--- a/libmate-desktop/mate-colorsel.c
+++ b/libmate-desktop/mate-colorsel.c
@@ -1025,15 +1025,15 @@ palette_paint (GtkWidget    *drawing_area,
 {
   gint focus_width;
   GtkAllocation allocation;
-  GdkRGBA bg;
+  gdouble color[4];
 
   if (gtk_widget_get_window (drawing_area) == NULL)
     return;
 
   gtk_widget_get_allocation (drawing_area, &allocation);
 
-  gtk_style_context_get_color (gtk_widget_get_style_context(drawing_area),GTK_STATE_FLAG_NORMAL, &bg);
-  gdk_cairo_set_source_rgba (cr, &bg);
+  palette_get_color (drawing_area, color);
+  cairo_set_source_rgba (cr, color[0], color[1], color[2], color[3]);
   cairo_rectangle (cr, 0, 0, allocation.width, allocation.height);
   cairo_fill (cr);
 
@@ -1221,26 +1221,6 @@ palette_change_color (GtkWidget         *drawing_area,
   g_free (current_colors);
 }
 
-static void
-override_background_color (GtkWidget *widget,
-                           GdkRGBA   *rgba)
-{
-  gchar          *css;
-  GtkCssProvider *provider;
-
-  provider = gtk_css_provider_new ();
-
-  css = g_strdup_printf ("* { color: %s;}",
-                         gdk_rgba_to_string (rgba));
-  gtk_css_provider_load_from_data (provider, css, -1, NULL);
-  g_free (css);
-
-  gtk_style_context_add_provider (gtk_widget_get_style_context (widget),
-                                  GTK_STYLE_PROVIDER (provider),
-                                  GTK_STYLE_PROVIDER_PRIORITY_APPLICATION);
-  g_object_unref (provider);
-}
-
 /* Changes the view color */
 static void
 palette_set_color (GtkWidget         *drawing_area,
@@ -1249,14 +1229,6 @@ palette_set_color (GtkWidget         *drawing_area,
 {
   gpointer pointer;
   gdouble *new_color = g_new (double, 4);
-  GdkRGBA  box_color;
-
-  box_color.red = color[0];
-  box_color.green = color[1];
-  box_color.blue = color[2];
-  box_color.alpha = 1;
-
-  override_background_color (drawing_area, &box_color);
 
   pointer = g_object_get_data (G_OBJECT (drawing_area), "color_set");
   if (!pointer || GPOINTER_TO_INT (pointer) == 0)
```

libmate-desktop/mate-colorsel.c
```diff
@@ -1162,19 +1167,18 @@ palette_change_color (GtkWidget         *drawing_area,
...
 
   priv = colorsel->private_data;
 
-  gdk_color.red = UNSCALE (color[0]);
-  gdk_color.green = UNSCALE (color[1]);
-  gdk_color.blue = UNSCALE (color[2]);
-  gdk_color.pixel = 0;
+  gdk_color.red = color[0];
+  gdk_color.green = color[1];
+  gdk_color.blue = color[2];
```
cwendling, line 1181: do you need to initialize `gdk_color.alpha`?

zhuyaliang: No need to initialize ```alpha```, this value is not needed
The ```gdk_color``` is ultimately used by ```mate_color_selection_palette_to_string``` and does not require parsing of alpha

libmate-desktop/mate-colorsel.c
zhuyaliang: ```#``` prefix needs to be preserved

libmate-desktop/mate-colorsel.c
zhuyaliang: Thank you for simplifying the code

libmate-desktop/mate-colorsel.c
zhuyaliang: Thank you review code

lukefromdc:
Any more updates on this?

libmate-desktop/mate-colorsel.c
cwendling: You mean the value in parses *has* to be formatted in a way compatible with `gdk_rgba_parse()`?

libmate-desktop/mate-colorsel.c
```diff
@@ -1371,9 +1315,14 @@ do_popup (MateColorSelection *colorsel,
...
-                  popup_position_func, drawing_area,
-                  3, timestamp);
+  if (trigger_event && gdk_event_triggers_context_menu (trigger_event))
+    gtk_menu_popup_at_pointer (GTK_MENU (menu), trigger_event);
+  else
+    gtk_menu_popup_at_widget (GTK_MENU (menu),
+                              drawing_area,
+                              GDK_GRAVITY_CENTER,
+                              GDK_GRAVITY_NORTH_WEST,
+                              trigger_event);
```
cwendling, line 1325: Unfortunately there seem to be a GTK bug here, if you use the keyboard to popup it appears off the window…
Using `gtk_widget_get_parent (drawing_area)` as the widget works, but it **is not the right solution**.

I'll try and see to report this upstream if I can understand something around it, unless anybody knows what's going on here?

libmate-desktop/mate-colorsel.c
```diff
@@ -1743,15 +1687,16 @@ key_press (GtkWidget   *invisible,
...
+  GdkDevice *device, *pointer_device;
   GdkDisplay *display = gtk_widget_get_display (invisible);
   GdkScreen *screen = gdk_event_get_screen ((GdkEvent *)event);
   guint state = event->state & gtk_accelerator_get_default_mod_mask ();
   gint x, y;
   gint dx, dy;
 
   device = get_device (display);
-  gdk_device_get_position (device, NULL, &x, &y);
+  pointer_device = gdk_device_get_associated_device (device);
```
cwendling, line 1698: That is *not* working, as `get_device()` already returns the display's seat's default pointer, [`gdk_device_get_associated_device()`](https://docs.gtk.org/gdk3/method.Device.get_associated_device.html) will give you the display's seat's default keyboard -- which doesn't have a position.

So either remove this and make `pointer_device = device`, or replace the `get_device()` call with `gdk_event_get_device((GdkEvent*) event)` so you effectively get the pointer associated with the keyboard triggering the event.

```diff
diff --git a/libmate-desktop/mate-colorsel.c b/libmate-desktop/mate-colorsel.c
index 44b6074..6bcd442 100644
--- a/libmate-desktop/mate-colorsel.c
+++ b/libmate-desktop/mate-colorsel.c
@@ -1567,17 +1567,6 @@ make_picker_cursor (GdkScreen *screen)
   return cursor;
 }
 
-static GdkDevice *get_device (GdkDisplay *display)
-{
-  GdkSeat   *seat;
-  GdkDevice *device;
-
-  seat = gdk_display_get_default_seat (display);
-  device = gdk_seat_get_pointer (seat);
-
-  return device;
-}
-
 static void
 grab_color_at_mouse (GdkScreen *screen,
 		     GdkDevice *device,
@@ -1688,13 +1677,12 @@ key_press (GtkWidget   *invisible,
            gpointer     data)
 {
   GdkDevice *device, *pointer_device;
-  GdkDisplay *display = gtk_widget_get_display (invisible);
   GdkScreen *screen = gdk_event_get_screen ((GdkEvent *)event);
   guint state = event->state & gtk_accelerator_get_default_mod_mask ();
   gint x, y;
   gint dx, dy;
 
-  device = get_device (display);
+  device = gdk_event_get_device ((GdkEvent *) event);
   pointer_device = gdk_device_get_associated_device (device);
   gdk_device_get_position (pointer_device, NULL, &x, &y);
 
```

libmate-desktop/mate-colorsel.c
```diff
@@ -2406,41 +2358,47 @@ mate_color_selection_set_has_palette (MateColorSelection *colorsel,
...
   if (priv->default_set == FALSE)
     {
       for (i = 0; i < COLORSEL_NUM_CHANNELS; i++)
-	priv->old_color[i] = priv->color[i];
+        priv->old_color[i] = priv->color[i];
     }
+
   priv->default_set = TRUE;
   update_color (colorsel);
+
```
cwendling, line 2401: Unnecessary blank line

```suggestion
```

cwendling (changes requested):

libmate-desktop/mate-colorsel.c
zhuyaliang: done

libmate-desktop/mate-colorsel.c
zhuyaliang: If you replace ```gtk_drawing_area_new()``` with ```gtk_button_new```, everything will be fine. I don't know if this is a bug in GTK, it will take time to verify

zhuyaliang: I have written test code, and this bug seems to only appear in ```gtk_drawing_area_new```
I plan to rollback the code

zhuyaliang: Using ```gtk_widget_get_parent (drawing_area) ``` is the correct way to handle the issue. I checked the ```GTK``` code and found the reason for the incorrect menu position，```gtk_menu_popup_at_widget``` is suitable for widgets that do not create windows for themselves.
For example, ```flowbox```, ```drawarea```, and ```eventbox``` use ```gtk_widget_set_has_window (widget, TRUE)```, which can cause the menu to pop up in the wrong position

zhuyaliang: The parent of ```drawarea``` is a ```frame```, so there is no problem using the parent
```
  frame = gtk_frame_new (NULL);
  custom_palette= palette_new (colorsel);
  gtk_container_add (GTK_CONTAINER (frame), custom_palette);
  gtk_grid_attach (GTK_GRID (grid), frame, i, j, 1, 1);
```

cwendling: > Using `gtk_widget_get_parent (drawing_area) ` is the correct way to handle the issue.

After research, in our case, it's probably the best solution indeed. It however only works because the parent is a frame that only contains the drawing area, so placing in the center of the frame is also the center of the drawing area; and even if we placed at the bottom-left it'd be only off by the borders of the frame (which might even be prettier).

But this is not a generic fix that would work in other situations though.  I just submitted the [issue upstream](https://gitlab.gnome.org/GNOME/gtk/-/issues/6166).
A usually better solution would be not having our own window on the widget, but GTK requires one to be a drag source, so we actually need one.

> I checked the `GTK` code and found the reason for the incorrect menu position，`gtk_menu_popup_at_widget` is suitable for widgets that do not create windows for themselves.

Yeah, it is basically broken if the widget's window has a non-zero position.

libmate-desktop/mate-colorsel.c
cwendling: I guess it's fine, so let's leave this alone.

libmate-desktop/mate-colorsel.c
```diff
@@ -1371,9 +1315,14 @@ do_popup (MateColorSelection *colorsel,
...
-                  popup_position_func, drawing_area,
-                  3, timestamp);
+  if (trigger_event && gdk_event_triggers_context_menu (trigger_event))
+    gtk_menu_popup_at_pointer (GTK_MENU (menu), trigger_event);
+  else
+    gtk_menu_popup_at_widget (GTK_MENU (menu),
+                              gtk_widget_get_parent (drawing_area),
+                              GDK_GRAVITY_CENTER,
+                              GDK_GRAVITY_NORTH_WEST,
+                              trigger_event);
```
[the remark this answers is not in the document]
zhuyaliang: Thank you for adding the explanation

cwendling (approved):
I didn't test every single features here I believe (some I don't know what they are exactly), but code seems OK and from what I tested it does seem to work nicely :+1:

lukefromdc:
Is this ready to go?

raveit65:
I tested mate-color-select with latest changes again and it works well.

PR #586, opened 2023-10-08
mate-rr-labeler: Fix compilation warnings use css replace gtk_widget_override_color

zhuyaliang:


lukefromdc (approved):
Builds with no warnings from this file.I take it this is used in the xrandr capplet? That capplet works fine with this installed. Please advise if this is incorrect and something else needs to be tested.

Nice reimplementation of something GTK has deprecated-we could potentially use it elsewhere in MATE by adding set_override_color to mate-rr-labeller.h though this would be a non-obvious place to find it

raveit65:
It is also used in `mate-settings-daemon/plugins/xrandr/msd-xrandr-manager.c:46:#include <libmate-desktop/mate-rr-labeler.h>
`
No idea, how to test it.

PR #587, opened 2023-10-09
wayland: fix uninitialized variable

lukefromdc:
We still have sn_display in the runtime checks to determine whether to call the startup notification, so we have to initialize it to NULL in the wayland case.


raveit65 (approved):
LGTM

PR #588, opened 2023-10-10
meson: Add meson support for static library

zhuyaliang:
Should add support for building static libraries in ```meson```

## test
```
meson build -Ddefault_library=both
```

### DIFF
--- a/libmate-desktop/meson.build
+++ b/libmate-desktop/meson.build
@@ -88,7 +88,7 @@ if cc.has_link_argument(test_ldflag)
   ldflags += test_ldflag
 endif
 
-libmate_desktop = shared_library(
+libmate_desktop = library(
   'mate-desktop-2',
   sources: sources,
   version: libversion,


lukefromdc:
Building with this does not stop mate-control-center (and possibly other builds) from erroring out with
`libtool:   error: cannot find the library '/usr/lib/x86_64-linux-gnu/libmate-desktop-2.la' or unhandled argument '/usr/lib/x86_64-linux-gnu/libmate-desktop-2.la'`
as this adds libmate-desktop-2.a to the build, but it is the .la file (a text file) being looked for by autotools

raveit65:
Should this be reviewed?
Auto-tools support both static and dynamic libs. Maybe this useful for meson too?

zhuyaliang:
```meson``` should support static library support,please test this PR

lukefromdc (approved):
This itself builds and runs fine. 

The issue I have trying to build anything with autotools  depending on anything built with meson failing on not finding .la files (not built in meson) might be local to my system. If not it probably has to be dealt with in the offending packages not here. I don't know enough about either build system to diagnoes that. GTK and Glib have been meson-only for years and most folks presumably don't have to hand pack manually edited .la files back into those packages

PR #602, opened 2024-02-28
MateImageMenuItem: Remove unnecessary label field

cwendling:
This was used in GtkImageMenuItem to support stock items, but this is a deprecated feature that didn't make it to MateImageMenuItem.  Now, this is entirely unused, and just leads to glue code maintaining an unused local copy of the label, effectively wasting a little bit of memory and adding complexity.

Just drop all references to it, without any functional difference.

See also https://github.com/mate-desktop/mate-panel/issues/1433#issuecomment-1969918905

### DIFF
--- a/libmate-desktop/mate-image-menu-item.c
+++ b/libmate-desktop/mate-image-menu-item.c
@@ -26,7 +26,6 @@ struct _MateImageMenuItem
 {
     GtkMenuItem  menu_item;
     GtkWidget   *image;
-    gchar       *label;
 };
 
 enum {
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
Builds and installs fine (test build was with meson). On restarting mate-panel, I found I could now drag icons for programs from the compact menu to the panel without the panel crashing, at least under wayland(x11 not tested yet). Previously doing this crashed the panel under wayland and x11 alike. On reverting to the 1.28.1 tarball build, the crash returned, proving this fixes it.

Some g_frees are gone because the items they freed are no longer created, that makes a double free impossible in those cases. Note that g_free just returns if called on a null pointer, wondering if that's whats stopping a crash with this code removed and a g_free call in mate-panel being called in a way that tries to free too much?

cwendling:
> Might be the chance fix mentioned in
> [mate-desktop/mate-panel#1433 (comment)](https://github.com/mate-desktop/mate-panel/issues/1433#issuecomment-1969918905) where MateImageMenuItem is now the same size as GtkImageMenuItem, if so we still need to fix mate-panel right

Exactly, we still need a proper fix for that issue.

zhuyaliang (approved):
Great changes

raveit65 (approved):
LGTM

lukefromdc:
We have a fix for the panel that works without relying on this, 
https://github.com/mate-desktop/mate-panel/pull/1434
and we already know removing the unused field doesn't seem to break anything elsewhere.

PR #647, opened 2025-05-20
GitHub action

yetist:
Try to use gitHub action


yetist:
The release test is here: https://github.com/yetist/mate-control-center/releases

lukefromdc (approved):
Again: I don't know the code here but the release test was just proven to work,

When we release 1.30 for real we'll want to deal with the soname issue that has been a blocker in Debian against picking up 1.28. Approving

mbkma (approved):

Issue #643, opened 2025-04-21
mate-color-select doesn't work (as expected)


linuxCowboy:
Saving a color in the palette does not work.

The tooltips say "drag-and-drop" or "right-click to save", but you can do neither.

**mate-desktop** 1.26.0 (_debian stable_)

(Yes, I reported it to _debian_, but they don't care...)

man page:
"It is not meant to be used on its own, but as a MATE Desktop helper application."

It's in the Desktop Applications Menu. So it's for users and should have full functionality. And why is it crippled?

Do I miss something?

lukefromdc:
The "copy" button works find, copies the color definition to the clipboard. I was not able to test drag and drop at the moment without closing my session as I am on the wayland session right now and that usually does not work on wayland with any application unless using portals.

copying it to the clipboard works, and removing it from the menu now will generate bug reports about why it is gone. No intentional feature limitation here, remember the code was originally developed decades ago as GNOME 2 and has been through the GTK2->GTK3 transition and many glib changes. It was originally used throughout MATE but has now been largely replaced by GTK 3's native color picker, which I consider to not have as good a UI but there was some kind of issue (maybe deprecations?) we had to deal with.

linuxCowboy:
Thank you for the explanations.

But why "removing from menu" - and not simply fix it? Store a value temporarily and set a color?

However, in _debian_ I can use the _qt5 color picker._

But the pure, essential _mate tool_ would like me better!

lukefromdc:
There are no plans to remove this from the menu. Note that we are a small team and this is of low importance. If anyone
wants to code up a drag and drop fix (e.g to drop into GIMP in x11), I can fire up an x11 session and test that.

linuxCowboy:
Hi Luke,

I was going back in time with **debian live** and **github**. Stefano Karapetsas brought this code in 2014. I planned to build it statically so I can use it today. But I was unable to build a _single functioning version_. "Save to palette" doesn't work!

I was back until _debian wheezy_ 2013. It seems too, that you can not build **mate-color-select** statically. `libtool -all-static` also not work. (I'm not sure about the last. Run `libtool` direct instead of `configure/make`. But since I couldn't find a good version anyway, I stopped.)

So it seems to me that the selector has never once worked as intended in _Debian_ over all the years - which I can hardly believe!

Please tell me I am wrong and _there is_ a good version?

lukefromdc:
[...]

PR #649, opened 2025-06-24
MATE Color Selection: Fix palette

linuxCowboy:
**MATE Color Selection** finally has a _working_ palette.

### DIFF
--- a/libmate-desktop/mate-colorsel.c
+++ b/libmate-desktop/mate-colorsel.c
@@ -43,6 +43,8 @@
 #define GTK_CUSTOM_PALETTE_WIDTH 9
 #define GTK_CUSTOM_PALETTE_HEIGHT 4
 
+#define SIZE_OF_COLOR_PALETTE (GTK_CUSTOM_PALETTE_WIDTH * GTK_CUSTOM_PALETTE_HEIGHT * 8)
+
 #define CUSTOM_PALETTE_ENTRY_WIDTH   20
 #define CUSTOM_PALETTE_ENTRY_HEIGHT  20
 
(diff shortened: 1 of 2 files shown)


linuxCowboy:
works in **Debian stable** with _mate-desktop 1.26.0_

linuxCowboy:
I wonder why you think it's working?

Yes, you see the _palette_. But you cannot change it! Not via **context menu** and not with **drag-and-drop**.

lukefromdc:
What exactly fails? I get a palette that shows, but that's all I am trying to do with it in this test

linuxCowboy:
Yes, I mentioned in an issue #643

lukefromdc (approved):
This does seem to work, though testing in wayland does not allow testing drag and drop to anything else due to the limitations of wayland

libmate-desktop/mate-colorsel.c
```diff
@@ -37,12 +37,14 @@
 #include "mate-colorsel.h"
 #include "mate-hsv.h"
 
-#define DEFAULT_COLOR_PALETTE "#ef2929:#fcaf3e:#fce94f:#8ae234:#729fcf:#ad7fa8:#e9b96e:#888a85:#eeeeec:#cc0000:#f57900:#edd400:#73d216:#3465a4:#75507b:#c17d11:#555753:#d3d7cf:#a40000:#ce5c00:#c4a000:#4e9a06:#204a87:#5c3566:#8f5902:#2e3436:#babdb6:#000000:#2e3436:#555753:#888a85:#babdb6:#d3d7cf:#eeeeec:#f3f3f3:#ffffff"
```
vkareh, line 40: Please don't change the default palette - it follows the Tango Desktop guidelines, which although it has not been in use by fd.o for a few years, are still part of MATE's visual heritage.

schemas/org.mate.interface.gschema.xml
```diff
@@ -75,6 +75,11 @@
       <summary>List of symbolic names and color equivalents</summary>
       <description>A '\n' separated list of "name:color" as defined by the 'gtk-color-scheme' setting</description>
     </key>
+    <key name="gtk-color-palette" type="s">
+      <default>'#FF0000:#FF2A00:#FF5500:#FF8000:#FFAA00:#FFD500:#FFFF00:#D4FF00:#AAFF00:#80FF00:#55FF00:#2AFF00:#00FF00:#00FF2A:#00FF55:#00FF80:#00FFAA:#00FFD4:#00FFFF:#00D4FF:#00AAFF:#0080FF:#0055FF:#002BFF:#0000FF:#2A00FF:#5500FF:#8000FF:#AA00FF:#D500FF:#FF00FF:#FF00D4:#FF00AA:#FF0080:#FF0055:#FF002B'</default>
```
vkareh, line 79: Same here

vkareh (changes requested):
Code looks good and works and the palette can be easily updated now, thanks!

schemas/org.mate.interface.gschema.xml
linuxCowboy: Take from my code whatever you like. My plan was only to make the **palette** working. This tool is in the _applications menu_ and should fully working!

I don't know anything about _wayland_ or which other tools use the _library_, but plan further rather heavy liftings of the **mate-colorsel*** stuff.

linuxCowboy: I made a single commit, not changing the palette. resolves #643

PR #651, opened 2025-06-27
Fix palette

linuxCowboy:
solves: mate-color-select doesn't work (as expected) #643


lukefromdc (approved):
This now functions as expected. Note that you can reset the palette back to its default values by selecting "use default" for org/mate/desktop/interface/gtk-color-palette in dconf-editor

lukefromdc:
Hoping we can get a second review

PR #690, opened 2026-05-04
mate-rr: Add hotplug_mode_update output property

vkareh:
Expose the RandR hotplug_mode_update property on MateRROutput so consumers can check whether an output supports dynamic mode updates on hotplug events (e.g. virtual machine guest resizing).

This is inspired by https://gitlab.gnome.org/GNOME/mutter/-/commit/957513242c26be458be7a101b83180e3f59f6a44 but made more generic so that m-s-d can handle the property and eventually fix mate-desktop/mate-settings-daemon#78

### DIFF
--- a/libmate-desktop/mate-rr-private.h
+++ b/libmate-desktop/mate-rr-private.h
@@ -47,6 +47,7 @@ struct MateRRScreenPrivate
     int				rr_minor_version;
 
     Atom                        connector_type_atom;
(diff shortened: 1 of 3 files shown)


mbkma (approved):
Clean addition. Exposing the RandR 'hotplug_mode_update' property through 'mate-rr' fits the existing API shape, and the X error handling around the property query looks in line with the rest of this code.

